PR #2, opened 2016-01-12
LSIO | Initial Plex changes 

pecigonzalo:


schakrava:
Thanks for the pull request @pecigonzalo! Since current image based Plex Rock-on is in use by a lot of users, it's best if your rockon is treated completely separate. So, I suggest adding a new file, say linuxserverio-plex.json with your recipe.

Once we are satisfied with your recipe, we can offer yours for all new users and deprecate the old one.

pecigonzalo:
Updated the files, does the custom_config at the container level work?

Cheers

pecigonzalo:
@schakrava new commits and rebase, once we have an initial working one, ill submit new features/modifications.

lsio-plex.json
```diff
@@ -0,0 +1,46 @@
...
+                },
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a Share for Plex configuration. Eg: create a Share called plex-config for this purpose alone.",
+                        "label": "Config Storage"
+                    },
+                    "/transcode": {
+                        "optional": true,
+                        "description": "Transcode directory to offload heavy writes in a docker container",
+                        "lable": "Transcode Location"
```
ericstewart22: @pecigonzalo "lable" should be "label". I'm still relatively new to Git so I figure you're probably better at making the change than I am.

pecigonzalo: Done

bskrtich:
I would recommend to not use 
```"opts": [
     "--net=host",
     ""
 ],```

It would be better to add all the needed ports.
```"ports": {
                    "32400": {
                        "description": "WebUI port. Suggested default: 32400",
                        "host_default": 32400,
                        "label": "WebUI port",
                        "ui": true
                    },
                    "32469": {
                        "description": "Plex DLNA Server port. Suggested default: 32469",
                        "host_default": 32469,
                        "label": "DLNA Server",
                        "ui": false
                    },
                    "5353": {
                        "description": "Bonjour discovery. Suggested default: 5353",
                        "host_default": 5353,
                        "label": "Bonjour discovery",
                        "protocol": "udp",
                        "ui": false
                    },
                    "1900": {
                        "description": "Plex DLNA Server UDP Port. Suggested default: 1900",
                        "host_default": 1900,
                        "label": "DLNA Server UDP Port",
                        "protocol": "udp",
                        "ui": false
                    }
                },```

phillxnet:
@bskrtich Thanks for this suggestion as it goes a previous version of the plex rock-on did use discrete ports, ie see change:
https://github.com/rockstor/rockon-registry/commit/dca3235dd672e63af2b0f45157fe44b61a3e0b9e
But there were discovery issues, however your change includes an additional port, 5353.
However when this was tried over in the following forum post:
https://forum.rockstor.com/t/plex-media-server/1321/13
It was found that the 5353 port clashed with Rockstor's avahi-daemon.
And in that thread I think @bug11 identifies the issue, hence the move to using the less desirable:
--net=host
setting.

PR #4, opened 2016-01-26
change share name example to transmission-data. Fixes #3

### LINKED ISSUE #3: improve transmission share name example
The transmission suggested share name is transmission-rockon where as this share is labelled in the GUI at selection time as Data Store which is what it is predominantly, that and the transmission config but this, presumably for ease of config, is not separated out.
I suggest that we change this to transmission-data to avoid the confustion that it stors the rock-on itself ie akin to:-
plex-data
btsync-data
owncloud-data
syncthing-data
note however that the we also have discourse-datastore and jenkins-home however these may be more relevant names to those projects than the generic data label used elsewhere.

This suggestion has been seconded by @Marinerdevil in the following forum post:-
http://forum.rockstor.com/t/transmission-rock-on-installation/965/4
Please update the referenced forum thread with development on this issue.

I am about to prepare a pull request to address this issue.
Note also that a parnter pr will be required to address the docs representation of this change. I will also attempt to get to this shortly as well.

phillxnet:
The suggested share name was previously transmission-rockon
which is confusing as it's only the data and config that uses this
share not the rockon itself which is stored on another specialized
share.

Fixes #3

### DIFF
--- a/transmission.json
+++ b/transmission.json
@@ -12,7 +12,7 @@
                 "launch_order": 1,
                 "ports": {
                     "51413": {
-                        "description": "Port used to share the file being downloaded. You may need to open it(protocol: tcp and udp) on your firewall. Suggested default: 51413.",
+                        "description": "Port used to share the file being downloaded. You may need to open it (protocol: tcp and udp) on your firewall. Suggested default: 51413.",
                         "host_default": 51413,
                         "label": "Sharing port"
                     },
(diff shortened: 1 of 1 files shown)


phillxnet:
@schakrava Ready for review.

schakrava:
Sorry for the delay @phillxnet. really appreciate the pr, thanks!

phillxnet:
@schakrava Cheers.

PR #8, opened 2016-02-12
Create sickbeard.json

dizzi90:
Modified BTSync.json to add Sickbeard support.


schakrava:
Thanks for the pull request @dizzi90, but I am afraid it doesn't work. First there were some parsing errors  in the json file and then the image itself(timhaak/docker-sickbeard:latest) is not available.

Could you please test and make necessary corrections?

PR #14, opened 2016-03-06
Modified btsync and added Crashplan with VNC and Squeezebox Server

henfri:


henfri:
Sorry Suman, I had to update the btsync again. 
Before, I was not able to test it due to a conflict with the existing btsync. Now I found a workaround and noticed that my previous version did not work.
This one is now the official btsync container and it does run.

schakrava:
@henfri I had to make a lot of changes to merge this pull request, but thanks for opening it. Next time, please remember to open one pr per rock-on so we don't have to stall all rock-ons for one problematic one. I couldn't merge Crashplan. In my testing, it proved too brittle and required a lot of tweaking to get it to work. Something is missing.

PR #16, opened 2016-03-13
Sonarr Rock-on created

tronder88:
initial testing works

### DIFF
--- a/couchpotato.json
+++ b/couchpotato.json
@@ -1,0 +1,59 @@
+{
+    "couchpotato": {
+        "containers": {
+            "couchpotato": {
+                "image": "linuxserver/couchpotato",
+                "launch_order": 1,
+                "opts": [
+                    [
+                        "--net=host",
+                        ""
+                    ]
+                ],
+                "ports": {
+                    "5050": {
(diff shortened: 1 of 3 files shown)


schakrava:
Hi @tronder88 ! Thanks for the pull request. I tested the couchpotato and it works. But one question, why do we need host networking (--net=host) option?

tronder88:
I have no idea, just used an existing JSON file and edited this, and since it works I have not tested without. Just a heads up: the SabNZB JSON have minor port issues.. Standard port for SABNZB is 8080, and i had to map host port 8085 for it to work, due to another JSON plugin which has reserved that port number.

wingrunr21:
You don't need the `--net=host` option on any of them. If the Plex JSON was copied, that'd probably be why they are there. Only thing I'd notice is that `linuxserver` recommends read-only access to `/etc/localtime`. Does Rockstor add that independently?

schakrava:
I've tested and cleaned up all three files. Thanks for your contribution @tronder88. 

@wingrunr21, regarding /etc/localtime mapping, I think it's a good idea to do that for all Rock-ons. Don't you think?

wingrunr21:
I can't think of a downside for most stuff.

Issue #20, opened 2016-03-23
couchpotato and sonar web links broken


phillxnet:
When clicking on the couchpotato 'name' link in the available Rock-ons page there is no external url but a relative one to the installed system, ie on a system named rockcube.lan we get:
```
https://rockcube.lan/<https://hub.docker.com/r/linuxserver/couchpotato></a>
```
The same issue exists with the sonar rock-on:
```
https://rockcube.lan/<https://hub.docker.com/r/linuxserver/sonarr></a>
```

tronder88:
Think issue can be closed now. I have informed Suman about the issue and he has corrected them.

phillxnet:
@tronder88 Thanks for the heads up.
Closing as fixed by #22 and https://github.com/rockstor/rockon-registry/commit/09f36ee530a1005129cbb7b02412ddde7e85d9a6

PR #22, opened 2016-03-24
Update couchpotato.json

tronder88:
Fix link? line 50.. local webadress is included before regular adress

### DIFF
--- a/couchpotato.json
+++ b/couchpotato.json
@@ -47,7 +47,7 @@
             "slug": ""
         },
         "volume_add_support": true,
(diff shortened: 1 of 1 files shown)


phillxnet:
@tronder88 is this a fix for #20 at least the couchpotato part, linking here just in case.

tronder88:
Yes, i think that will solve the link issue. Last commit with link without <> around the address.

schakrava:
Thanks @tronder88 !

PR #30, opened 2016-04-02
symform entry missing in root. Fixes #29

### LINKED ISSUE #29: symform entry missing in root.json
The symform.json Rock-on is in rockon-registry but there is no accompanying entry for the same in root.json. Adding under this issue to restore consistency.

phillxnet:
@schakrava Found this discrepancy so simple pr to address it. Not sure if I've missed a comment relating to this though.

Consequences as far as I can see are the non appearance of the recently added Symform Rock-on in available Rock-ons.

Not sure how to test these changes though without the repo update so submitting as untested.


schakrava:
Thanks @phillxnet

PR #32, opened 2016-04-05
Added Nethost option

tronder88:
Nethost option added back to plex, due to portconflict.
https://forum.rockstor.com/t/plex-media-server/1321/11   It is believed
that this will solve issue.


tronder88:
Implemented

PR #35, opened 2016-05-24
added MariaDB rock-on

dschoorisse:
MariaDB database, based on <https://hub.docker.com/r/linuxserver/mariadb/>

### DIFF
--- a/mariadb.json
+++ b/mariadb.json
@@ -1,0 +1,49 @@
+{
+    "MariaDB":{
+        "containers":{
+            "Sonarr":{
+                "image":"linuxserver/mariadb",
+                "launch_order":1,
+                "ports":{
+                    "3306":{
+                        "description":"MariaDB port. Suggested default: 3306",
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks for the pull request @dschoorisse. A couple of comments

1. The container is named Sonarr, please name it as linuxserver-mariadb
2. It would be very helpful to add a "more_info" with information such as how to access the db from cli etc.. See openvpn.json for example.

dschoorisse:
Oops, renamed it locally but forgot to check in. I will fix it. Thanks for you respons!

dschoorisse:
Hi, this is fixed now :)

schakrava:
I found an error in the naming of the database password env variable. Please see my changes. With the wrong env variable, root's password is not set, so you can login without password :)

Thanks for your work on this. I think this is a very useful Rock-on for advanced users wanting a mariadb/mysql database.

PR #37, opened 2016-06-01
added Headphones rock-on

dschoorisse:


schakrava:
Nicely done @dschoorisse. Just a minor correction I made is to remove the mapping of /etc/localtime. That is done by default for all rockons.

dschoorisse:
Thanks for merging and for your comment. Still learning :) Thanks!

PR #41, opened 2016-06-26
Add Unifi Controler

bskrtich:


schakrava:
Thanks @bskrtich . Quick question. Why do we need host networking? Looks like all necessary ports are mapped?

bskrtich:
Because there is a whole list of ports that need to be open for this service to work.
```
8080/tcp
8081/tcp
8443/tcp
8843/tcp
8880/tcp
3478/udp
```

schakrava:
You may know this already, but it's better to avoid --net=host for 1) it compromises security by letting the container open any low numbered port and 2) ports opened by the container may unintentionally conflict with some service running on the host or another container. I am not a docker expert by any definition, but I find this [doc](https://docs.docker.com/v1.8/articles/networking/#container-networking) useful.

I think in this case, it's more manageable long term to have those 6 ports mapped. I looked at the [image page](https://hub.docker.com/r/jacobalberty/unifi/) and they suggest --net=host for a "quick" setup. But they have nicely documented all necessary ports as you've listed above as well.

If you agree, could you please update the json file and 1) remove host networking and 2) add all 6 ports? I'd be happy to give it a try.

phillxnet:
@bskrtich Yes that's a lot of ports, and although it's unwieldy I think it's still preferred to the host networking.  Also looks like the problem port of 5353 isn't required so that's good and hopefully we won't get a repeat of the Plex Rock-on port issues ie in #2 and #42 when you initially went the preferred port only route.
Thanks for yet another contribution by the way.
I say try with all the ports mapped and drop the less desirable --net=host and if it works (unlike the Plex rock-on with that arrangement) then we are golden.

phillxnet:
@bskrtich and @schakrava Sorry looks like I was butting in here.

bskrtich:
I have made the changes you requested.

schakrava:
Thank you @bskrtich! I merged your changes finally. Not sure why it's not showing merged here on github, but we are set.

PR #42, opened 2016-06-26
Fix some formatting for plex.json

bskrtich:
Also fix some formatting.

See https://github.com/rockstor/rockstor-core/issues/1209

### DIFF
--- a/plex.json
+++ b/plex.json
@@ -3,13 +3,13 @@
         "containers": {
             "plex-linuxserver.io": {
                 "image": "linuxserver/plex",
-                "launch_order": 1,
                 "opts": [
                     [
                         "--net=host",
                         ""
                     ]
                 ],
+                "launch_order": 1,
                 "ports": {
                     "32400": {
                         "description": "WebUI port. Suggested default: 32400",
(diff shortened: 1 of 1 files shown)


phillxnet:
@schakrava @bskrtich pleases note my comments in:
https://github.com/rockstor/rockon-registry/pull/2
Replicated here for ease:

"@bskrtich Thanks for this suggestion as it goes a previous version of the plex rock-on did use discrete ports, ie see change:
https://github.com/rockstor/rockon-registry/commit/dca3235dd672e63af2b0f45157fe44b61a3e0b9e
But there were discovery issues, however your change includes an additional port, 5353.
However when this was tried over in the following forum post:
https://forum.rockstor.com/t/plex-media-server/1321/13
It was found that the 5353 port clashed with Rockstor's avahi-daemon.
And in that thread I think @bug11 identifies the issue, hence the move to using the less desirable:
--net=host
setting.
"
So not sure we can use this arrangement without first addressing this 5353 port clash.

bskrtich:
Ok I changed it to just fix a bit of the formatting with the json file.

schakrava:
Thank you @bskrtich !

PR #45, opened 2016-06-30
Daemon Port

TimothyVandenbrande:
allow access to the daemon,
since if you can't automate these actions and need to do everything through the web-ui, it just makes the whole thing a bit obsolete.

### DIFF
--- a/deluge.json
+++ b/deluge.json
@@ -4,14 +4,6 @@
             "Deluge": {
                 "image": "linuxserver/deluge",
                 "launch_order": 1,
-		"opts": [
-                    [
-                    	"--net=host",
-                        "",
-			"-v",
-			"/dev/rtc:/dev/rtc:ro"
-		    ]
-		],
                 "ports": {
                     "8112": {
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @TimothyVandenbrande 

I noticed that we are using host networking (--net=host). In presence of that port mapping doesn't really make sense beyond telling UI which port to take the user to.

Do you think we can remove host networking in favour of specifically mapping the required ports, presumably only two(8112 and 58846)?

Also I don't see a need to map /dev/rtc as every rock-on automatically gets /etc/localtime mapped.

Do these changes make sense to you? If so, please make them and I'd be happy to review them.

TimothyVandenbrande:
I updated your request. It is true, it didn't really make a lot of sense, but tbh, I never really checked it.
Now, I have cleaned out the file, also removed all tabs and stuff so if you still find something else that I looked over, feel free to let me know, otherwise, this is it.

schakrava:
Thank you @TimothyVandenbrande !

PR #46, opened 2016-06-30
Add Rocket.Chat rock-on

lowsky:
Adding [Rocket.Chat](https://rocket.chat/) rock-on.

this was my first rock-on, so I appreciate any hint for improvement. Eg.
- I was not fully sure, if I specified the linking of both Docker container correctly

Thanks for creating RockStore!

### DIFF
--- a/rocketchat.json
+++ b/rocketchat.json
@@ -1,0 +1,57 @@
+{
+    "Rocket.Chat": {
+        "container_links": {
+            "Rocket.Chat": [{
+                "name": "mongodb.rocketchat",
+                "source_container": "mongodb.rocketchat"
+            }]
+        },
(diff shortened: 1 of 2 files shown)


lowsky:
Hi @schakrava,

I found and fixed a bug in the configuration, which I didn't notice before,
and upgraded the linked docker images.

Can you check it?

Installation should be straight forward...

Thanks, Robert

schakrava:
This is really cool @lowsky. Thank you!

I am planning to use this in the spirit of dogfooding. Have you configured HTTPS? It would be good to know how. I don't know much about meteor apps, wondering if a cert can be added easily or the only way to do so is with something like an nginx reverse proxy.

PR #48, opened 2016-07-05
Add subsonic rockon config

jfearon:
Add rock on config for subsonic (http://www.subsonic.org/pages/index.jsp) using https://hub.docker.com/r/hurricane/docker-subsonic/

### DIFF
--- a/sickrage.json
+++ b/sickrage.json
@@ -9,7 +9,7 @@
                         "description": "Port for Sickrage Web interface. Suggested default: 8081.",
                         "host_default": 8081,
                         "label": "WebUI port",
-                        "procotol": "tcp",
+                        "protocol": "tcp",
                         "ui": true
                     }
                 },


schakrava:
Thanks for your contribution @jfearon! Subsonic is pretty cool and would be nice to have as a Rock-on. After reviewing your changes, I have a couple of questions.

1. What made you choose the hurrican/subsonic image over others? It seems there are several choices without a clear winner in terms of official-ness or number of downloads etc..

2. Have your tried without the TZ env variable? /etc/localtime is mapped for all containers so the timezone is consistent with the Rockstor system. Given that, I am wondering if that env variable is required.

jfearon:
Hi,

1.  The most popular docker image (the 100k plus akegata one) is only subsonic 5. Once that was out of the way I went for number 2 - the config options and mount points look like what i wanted.  Having said that the next one down (hurricane/subsonic not hurricane/docker-subsonic) looks like it has been more recently pushed.  It offers a bit more config too - you can give a user/group id (like the plex plugin) and a playlists share seperately.  This probably isn't a very satisfactory answer - I don't have a strong preference - the other hurricane image is more configurable, but more complex.  Everything below the top 3 was either an older version of subsonic, or was just one mount point - music - rather than the separate music / podcasts.

2. I haven't tried this - I'll give it a go, and remove from the config if it's not needed.

schakrava:
@jfearon, I've removed TZ part and merged it. Thanks for your contribution! I am looking forward to using Subsonic.

PR #49, opened 2016-07-21
Pull for fork - add PocketMine

magicalyak:
Adding a rock-on for using minecraft server on Rockstor.  the Minecraft server is the pocket edition which works on iphone, kindles and other mobile devices.  Using the rockon allows you to control the service through rockstor. I had to add the restart option because it always takes twice to get this up but it should be solid and should work great!


pocketmine.json
```diff
@@ -0,0 +1,36 @@
+{
+    "PocketMine": {
+        "containers": {
+            "pocketmine": {
+                "image": "magicalyak/pocketmine",
+                "launch_order": 1,
+                "opts": [
+		  [
+			"-it",
+			"--restart=always",
```
schakrava: Thank you @magicalyak . I have two questions here. 

1. Why ```-it```?

2. The rock-on framework automatically adds ```--restart=unless-stopped```, not sure if you've noticed it. Given that, any reason, why we'd want ```--restart=always```?

magicalyak: -it should be removed it's not needed and I need to set resets to 1 or 2 it always needs one restart. Will fix soon

magicalyak: Done

magicalyak:
This definitely should be tested by someone else.  Pocketmine seems a little unstable at times (not this, but the program itself).  Plus it's a manual update to getting this to work.

magicalyak:
> @magicalyak Hello yet again. I know this is way old but given the image may have improved in stability, assuming auto update, things may be better now.
> 
> Let me know what you want to do, otherwise I think this one is slated for closure soon.

This is now updated to the latest bedrock edition that just came out, they rewrote the server and it doesn't need restart anymore so that's removed.  Just need to test this once and should be good to merge (unless someone else wants to test).

magicalyak:
Sounds good, I must have synced the whole dir.  let's do a new PR

Issue #44, opened 2016-06-28
Create a Zoneminder plugin


zorrobyte:
After purchasing several IP cameras and running Synology's Surveillance Add-on, I'd like to bring some sort of IP Camera DVR to Rockstor. Zoneminder is a free and open source solution for this. I'll see what I can write up, pending a successful Rockstor install.

phillxnet:
@zorrobyte This would be great, but last I looked although there was an official zoneminder docker image it wasn't fully tested. So at least back then it was early days. Thanks for you efforts on this one.
Linking to a prior mention / request (by me actually) and seconded by another entry in the forum:
https://forum.rockstor.com/t/rock-on-requests/41/50

And linking to the official docker hub image for convenience:
https://hub.docker.com/r/kylejohnson/zoneminder/
N.B. Under the Docker subheading there they have:
"ZoneMinder, and the ZM webserver, will run using the Dockerfile contained in this repository. However, there is still work needed to ensure that the main ZM features work properly and are documented."
It may be that only locally connected devices are compromised and IP cameras may well work. If that is the case it could be included as a caveat in the Rock-on user facing info.

And their main GitHub page:
https://github.com/ZoneMinder/ZoneMinder

phillxnet:
Linking to another forum post mentioning zoneminder as a request:
https://forum.rockstor.com/t/new-user-migrating-from-freenas/1583

magicalyak:
Does this still need to be done? I can try to take a crack at it?

phillxnet:
@magicalyak Not as far as I'm aware. @zorrobyte showed interest earlier on but I haven't seen any commits against the issue. Thanks for stepping up. I did look the other day on docker hub and there does appear to be a good few images, I think it would be best if we could use the official one as I indicated earlier though. If it works that is. If you make any progress do please note it here in case anyone else wants to pick up where you left off. Unless of course you manage to sort it.

Thanks.

I see you have already put in a pr for PocketMine / Minecraft server; nice idea.

magicalyak:
Let me see what I can do, this shouldn't be too difficult and I do feel a lot more confident after getting pocketmine to work (it had a LOT of quirks and no one else had it working on the new API).

magicalyak:
It's pending a pull request
[https://github.com/rockstor/rockon-registry/pull/50](url)

magicalyak:
[...]

PR #50, opened 2016-07-21
Zoneminder Rockon Addition

magicalyak:
This is the v1.29 of aptalca's docker image (using the zoneminder package from phusion/baseimage).  This is preferrable than compiling from zoneminder from scratch (even says so on the readme).  There are over 10K of pulls for this version and I tested it on my own rockon machine.  One big caveat is it needs to run at privileged user once (this is included) and it will change the shm to 4096m from the default 64m.  This is necessary for zoneminder to work.  The latest docker has a runtime flag one could use but our docker version is 0.9 (and it requires 0.11)


magicalyak:
The WebUI requires you add /zm after it.  Not sure how to do that (is it a slug command?).  For example it will take you to http://rockstor:8080 but you need to add http://rockstor:8080/zm

magicalyak:
Fixed the URL, it's all good now 💯

phillxnet:
@magicalyak That was quick. Well done and thanks. A note on mapped volumes, I see a config mapping but is it possible to map a volume for all the thousands of images that zoneminder builds up? I know they are planning on moving to mp4 or the like but when I last set this up it generated a tone of jpg's I think it was. It might be good to have that storage survive a Rock-on removal and re-install process. I'm assuming the db is mapped to the config volume as well for the same reasons. Ie so we end up with 2 volumes, a config and a data. Sorry if I've missed something here.

Well done on the URL fix. I expect @schakrava will be popping in on this one soon.

magicalyak:
You can add volumes after the install, just stop the zoneminder rockon and click setup and add volume.  I originally thought about adding it but honestly, it's a pain if you just wanted one share to do all the work.  Once you add it (let's say it's /data) then just point zm config to it.  Once 1.30 is stable I can add it (I may rewrite the dockerfile but one used is pretty solid and heavily used).

magicalyak:
let me add /data standby

phillxnet:
@magicalyak OK, so as is all data, ie config, db, history of captured events (the thousands of images) is all stored in the config volume then? And it's possible to stop the Rock-on, uninstall it, re-install it and have persistence of all settings and previously captured events (assuming the same volume is mapped to the same share of course)?

In my opinion this is a great addition to Rockstor by the way.

magicalyak:
I'm double-checking, I understand your ask but let me double check to make sure.  It is an all-in-one (DB included, etc).

phillxnet:
@magicalyak I only mention it as most of our other Rock-ons split the config / data elements as it allows folks more freedom to split these 2. Of course this is a pretty complex rock-on. Ie mapping between db and images and config. I'm afraid it's year since I set one up so just can't remember. Your call on if it's not feasable. Ie it may just be better to keep all the config + db + past event all together to maintain their integrity.

Sorry I feel like I'm interfering here. Let us know what you think. I just remember many many thousands of images, ie one for each frame of captured events.

magicalyak:
Last tweak, just to put the warning on the description about privileged flag.
the /config is the only one needed, it the place for mysql, config settings and the data.  I would think splitting them would be ideal also but zoneminder has a strict structure and we'd need to mount the new drive inside the old.  I suppose we could make /data point to /config/data but I'd have to do more testing.  Definitely works well as it stands but if I can split, I'll let you know.

magicalyak:
I segmented the mysql but the data is having permissions issues...

magicalyak:
Not sure it's worth only separating the mysql but I can if desired.  I can't separate the data/config, it's bundled somehow and I'd have to rewrite the dockerfile (which I could but then you're totally relying on my docker hub version...)

magicalyak:
Ok db separated but data is not, permissions issues when I tried.

magicalyak:
images and events are their own shares, so I've got 4 shares required now
mysql
config
images
events

phillxnet:
@magicalyak OK, so be it. As you say there is also an upside of simplicity. I wonder if it's worth re-labelling the Config as Config & Data.

I see that you have just split off the db, If that works then it would allow people to cater for the CoW db nature and configure accordingly, ie more frequent defrags etc.

Thanks for looking into and trying out all of this by the way. I think this one is now over to @schakrava for final review. Are you happy that it's OK' as is' or would you rather keep them all together?

Our original Owncloud Rock-on has a separate share for it's db so we are consistent with that, but on the other hand our new Owncloud-official has an all in one labelled just "Storage". Akin to your original.

I'll leave this up to you and stop meddling.

Oh I see you have taken this splitting of shares to heart.

My thoughts are that shares that have a different use pattern are important to split off (if feasible); too many and we move away from keeping things simple and we become unwieldy. DB activity is hard on CoW file systems as it causes massive fragmentation. Also in a security related Rock-on appliance such as zoneminder there are maybe additional concerns where people may want to archive some elements, or put some on faster storage.

Your preferences and those of @schakrava and @MFlyer now I think.

I'm quite chuffed by this Rock-on addition as it goes and thanks for being so accommodating.

magicalyak:
I'm more towards the 3 commit and leaving the config and data together.  You can add storage to /config/data/events later if you like or you can leave it alone.  I'm realizing that my latest didn't create the sounds directory so I'm going to commit once more backing out the complexity.  As it stands there will be DB and CONFIG but you can add more shares later.

magicalyak:
This should be final now and good to go.

phillxnet:
@schakrava is merge master in these parts so I'll bow out. Well done and I agree with your final shares split. Db's are known to be tough on CoW so at least we cover that.

Thanks for all your efforts on this one.

magicalyak:
@schakrava I removed the restart line on this one, otherwise should be good!

phillxnet:
@magicalyak Hello again. Just had another quick look and noticed that there are 2 "volume_add_support" entries, Presumably you're intention was to only have the second one. Also there appears to be a mix of tab chars and spaces. My understanding is that we use spaces (in sets of 4). Although we do still have a mix hear and there but we are working on that :).

magicalyak:
Understood I'll fix today and update

phillxnet:
@magicalyak Thanks, but no rush. In your own time and just let us know. It may be some time before @schakrava does another round of merges anyway. Just thought I'd make a note as I was passing again. I know you've tested everything so it's just an aesthetic really.

magicalyak:
This is complete

zoneminder.json
```diff
@@ -0,0 +1,45 @@
...
+                "opts": [
+                         [
+                          "--privileged=true",
+                          ""
+                          ]
+                         ],
+                "ports": {
+                    "80": {
+                        "description": "Port for running ZoneMinder. You shouldn't open this externally (it's unsecure). Do not use 80 (it will interfere with RockStor)",
+                        "host_default": 8080,
```
schakrava: Thanks for this note, but happy to say that as of 3.8-14.02, Rockstor service(via nginx) no longer uses port 80. Let's leave this note here, however till we release 3.8-15.

schakrava:
Nice work @magicalyak . I am going to try and set it up at home. Also, thanks @phillxnet for pitching in!

I've been testing docker 0.11. Will make it available in testing repo soon. The privileged mode is a concern, but thanks for mitigating it by clearly stating it in the description.

magicalyak:
This issue can be closed also (or did that need to wait for the full release)?
https://github.com/rockstor/rockon-registry/issues/44

phillxnet:
@magicalyak I've just put a little note on #44 to this effect. As the json is now in the repository it's available to all Rockstor installs upon them pressing the  "Update" button on the Rock-ons page so it's already released as such. I often put a "Fixes #issuenumber" in a pr so then the referenced issue gets closed automatically by @schakrava's merge.

PR #51, opened 2016-07-21
ghost blog writer

magicalyak:
Ghost rock-on for writing blogs, super lightweight and from official sources (docker)


schakrava:
This is really neat. Thanks @magicalyak

PR #52, opened 2016-07-30
U torrent Rock-on

Mahmoud87:
**uTorrent Rock-on**

Created the Rock-on using [dbarton](https://hub.docker.com/r/ekho/utorrent/) docker image. Thanks to @phillxnet for all of his help to get this working.
Please if you locking to create your own Rock-on check this [thread](https://forum.rockstor.com/t/request-rockon-utorrent/1824/15)

### DIFF
--- a/uTorrent.json
+++ b/uTorrent.json
@@ -1,0 +1,45 @@
+{
+	"utorrent": {
+		"containers": {
+			"utorrent": {
+				"image": "dbarton/utorrent",
+				"launch_order": 1,
+				"ports": {
+					"8080": {
+						"description": "Port for utorrent Web interface. Suggested default: 8080.",
+						"host_default": 8080,
+						"label": "WebUI port",
+						"procotol": "tcp",
+						"ui": true
+					}
+				},
+				"6881": {
+					"description": "Port for incoming data. You may need to open it(protocol: udp) on your firewall. Suggested default: 6881. Should NOT be changed",
+					"host_default": 6881,
+					"label": "uTorrent port",
+					"procotol": "udp",
+					"ui": true
+				},
+				"volumes": {
+					"/media": {
+						"description": "Directory for your downloaded media.",
+						"label": "Downloaded Data"
+					},
+					"/settings": {
+						"description": "Holds all the utorrent settings files and databases.",
+						"label": "uTorrent Settings"
+					}
+				}
+			}
+
+		},
+		"description": "BitTorrent client by [dbarton and Mahmoud87]",
+		"more_info": "<h4>uTorrent WebUI Logins.</h4><p>Username: admin / Password: (Leave it blank). You can always change the logins after your first sign in, go to settings >> WebUI.</p>",
+		"ui": {
+			"slug": "gui"
+		},
+		"volume_add_support": true,
+		"website": "https://hub.docker.com/r/ekho/utorrent/",
+		"version": "alpha"
+	}
+}
--- a/uTorrent.json
+++ b/uTorrent.json
@@ -39,7 +39,7 @@
 			"slug": "gui"
 		},
 		"volume_add_support": true,
-		"website": "https://hub.docker.com/r/ekho/utorrent/",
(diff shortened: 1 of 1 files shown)


phillxnet:
@Mahmoud87 Thanks for the contribution and attribution and well done on getting this far so quickly.

Just a quick note on polish / presentation: in your description in this pr you link to ekho's docker image rather than dbarton's. Probably best to edit this to avoid confusion during the review period. Another improvement you could do to the json file itself is to replace all your 'tab' characters with collections of 4 spaces. Various projects have differing standards on how stuff is indented and although we have all accidentally used 'tab' chars there is an ongoing concerted effort to use only spaces. Most files use collections of 4 spaces. Bit of a drag but helps with understanding the file structure as spaces are always just that (one space) but 'tab' characters are a little more meta and can change from one presentation method to another. This also helps with presenting things that are at the same level in the json as at the same indentation. I.e. the 2 port entries look in the GitHub viewer to be at different levels but they are supposed to share a level; that of being entries inside the "ports": field (see further notes on this below). The json file format itself doesn't care but it makes it easier to work on / work out.

On the ports entry, it looks like you have labelled both ports as pertaining to the ui with the  **"ui": true** on both ports. Only the 8080 port should have this. By the looks of it the 6881 port entry is not quite at the correct level: relating to indentation problem; i.e. each open "{" has a partner close "}" and ports is closed before we get to the 6881 entry. So the "6881" entry is actual at a 'level' higher than ports so is effectively outside of the ports section and so has no effect.

Hope that makes sense, and to keep a track of your various changes/commits you can name / label each commit, ie once you change the tabs to spaces you could label that commit accordingly, and once the 6881 is moved 'inside' the ports section you could label that as another change. This really helps with seeing what has been done to a file over time and often why. I suspect that during install only the 8080 port option was presented as configurable as that was the only known entry 'inside' of the ports section. As is the '6881' entry is at the same 'level' or depth as the 'image', 'ports', and 'volumes' section and given it is not a known entry it's probably just ignored by Rockstor's parser.

Hope that helps.

By the way just setting your editor to use 4 spaces instead of tabs and then doing an auto format (if the editor is capable of understanding json files) is often all that is needed to see what thing is at what level and to get a whole sale tab character removal at the same time.

Mahmoud87:
Hi Phil, sorry mate for the late reply had a busy weekend finally my RockStor went live, I was testing all this time on a VM, ran into some issue when it went live but good news all sorted :100: so far. I see what you saying about "6881" port, I was so confused as well during the installation it never asked me to expose this port and wondered what it's use then I just considered it redundant (noob move) lol. I can confirm the port "6881" issue has now been fixed, I have found out this port is the Incoming TCP Port in the uTorrent app, Also I have removed the "ui" = true and left it only for the port "8080". I will keep in mind when I change anything on git to label it to be more informative with the change I have made. Please check these pictures of the [uTorrent Rock-on installation](https://drive.google.com/open?id=0B_MlIZwAPDEXcjhseGNmbUdFVXc) and I will update the .json file little later

Mahmoud87:
@phillxnet All done man if you don't mind can you have a look at the file :) thanks mate

phillxnet:
@Mahmoud87  That looks great, well done; and nice detailed commit message there :).
Oh and nice catch on the "procotol" to "protocol"; I hadn't seen that. Thanks for your efforts on this and currently it looks good to me. If you have tested the install with the file as is and the html stuff shows up OK on the information page then it's your call to state if you think this is ready for final review and hopefully merging by @schakrava .

Mahmoud87:
Yeah @phillxnet it's all ready now, I have been using it for the last few days, no problems at all. All the HTML stuff showing perfectly fine. Yup I think it's time for review and hopefully @schakrava will merge it. Thanks again mate for all your support 👍

schakrava:
Nice work @Mahmoud87 and thank you @phillxnet for your input. Sorry it took me a while to get to this, but happy to merge it!

Mahmoud87:
Thanks @schakrava and @phillxnet glad I could add something to Rockstor community.

PR #57, opened 2016-10-22
add haproxy-letsencrypt rock-on (haproxy with letsencrypt integration)

daniel-illi:
This is a haproxy rock-on with integrated letsencrypt client.
On startup, a certificate for all configured hostnames (env variable) is requested from letsencrypt and all 3 weeks the certificates get automatically renewed.

I use this rock-on to make all my configured services (rock-ons) available over https on the standard port 443 and with valid certificates.

As the port 443 is already taken by the rockstor web interface, I run the haproxy on a different port (i.e. 1443) and configured my router to forward internet requests on port 443 to the haproxy port on the rockstor host.

### DIFF
--- a/haproxy-letsencrypt.json
+++ b/haproxy-letsencrypt.json
@@ -1,0 +1,55 @@
+{
+    "haproxy-letsencrypt": {
+        "description": "Reliable, High Performance TCP/HTTP Load Balancer with letsencrypt integration.",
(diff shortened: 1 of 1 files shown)


schakrava:
This is pretty cool. I'll personally use it. Thanks @daniel-illi !

PR #59, opened 2016-11-22
Added home-assistant.io to registry

KrayZee:
Home automation software written in Python3 (https://home-assistant.io/)
Supports many Smart Home hardware and software including Philips Hue, Google Chromecast, Plex, TP-Link Smart Plug, Vera, and many others.


schakrava:
Hi @KrayZee . Thanks for introducing homeassistant. I did not know it existed. It looks pretty cool and I like that it's written in Python. Could you address my review comments?

KrayZee:
Hi @schakrava 
I'd like to, but it looks I can't find any review comments.

phillxnet:
@KrayZee Hello, just jumping in on this one as I also can't see any comments but pretty sure I saw some recently from @schakrava  relating to the redundancy of mapping /etc/localtime:
A similar comment re /etc/localtime is available as item 2 (13th July) here:
https://github.com/rockstor/rockon-registry/pull/48
and the other comment I seem to remember being about the --net=host option which we really don't want to use if we can at all help it. It was found that without this the plex rock-on just failed to work but many others are running without it, best if we can avoid it really.

I've found a similar @schakrava comment re --net=host here:
https://github.com/rockstor/rockon-registry/pull/45

Hope that helps.

schakrava:
ok, my comments are visible now. I forgot to press a button, sorry.

home-assistant.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    "website": "https://home-assistant.io/",
+    "version": "1.0",
+    "containers": {
+      "home-assistant": {
+        "image": "homeassistant/home-assistant",
+        "launch_order": 1,
+        "opts": [
+          [
+            "--net=host",
```
KrayZee: according to documentation for home-assistant:
"Home Assistant must be on the same network as the devices for uPnP discovery to work. If running Home Assistant in a Docker container use switch --net=host to put it on the host’s network."
So removing this option will break discovery for devices such as Google Chromecast, Philips Hue, Plex Mediaserver, etc. I tried it without --net=host, rock-on starts, but discovery doesn't work indeed. So if it is possible i think it's better to keep this option

KrayZee:
redundant mapping /etc/localtime is removed. Also --net=host seems to be necessary for application to work due to UPnP network discovery

home-assistant.json
```diff
@@ -0,0 +1,40 @@
...
+            "",
+            "-v",
+            "/etc/localtime:/etc/localtime:ro"
```
KrayZee: done

schakrava:
Thank you @KrayZee !

PR #61, opened 2016-12-02
Add Collabora CODE rock-on

daniel-illi:
Collabora Online Development Edition (CODE) is an online office suite based on libreoffice which can be *integrated with owncloud and nextcloud*. It allows viewing and editing office documents in all popular office file formats, including text documents (docx, doc, odt,…), spreadsheets (xlsx, xls, ods,…) and presentations (pptx, ppt, odp,…).

This rock-on uses the official docker image available here:
https://www.collaboraoffice.com/code/
https://hub.docker.com/r/collabora/code/

### DIFF
--- a/collabra-online.json
+++ b/collabra-online.json
@@ -1,0 +1,39 @@
+{
+    "collabora-online": {
+        "description": "Collabora Online is a LibreOffice-based online office suite that can be integrated with owncloud/nextcloud.",
+        "version": "1.0.0",
+        "website": "https://www.collaboraoffice.com/code/",
+        "more_info": "",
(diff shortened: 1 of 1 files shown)


schakrava:
Thank you @daniel-illi . I couldn't get it to work on my vm, but looking forward to trying it out on a different system with better resources. But it installs fine and looks good. Do note that I renamed the file and also removed --restart=always. --restart options is added by the rock-on logic for all containers, so no need to add it.

daniel-illi:
Thanks for the feedback @schakrava and thanks for merging!

PR #68, opened 2016-12-14
Added Jackett

porkie02:


schakrava:
Thanks @porkie02 . I just had to correct the image name, but everything else looked good. You've added the tag to the image name, where as  it should be it's own key,value pair. or if it's latest, you can just leave it. Rockstor will assume the tag to be latest.

PR #70, opened 2017-01-16
Duplicati2 canary

dschoorisse:
Tested on 3.8.16-1 and 3.8-14.22.

Only a share for config storage is mandatory. Extra shares have to be created by the user (folder to backup, or to write backups to).


schakrava:
Thanks @dschoorisse. It works great. And I discovered a minor bug in our rock-ons logic while testing duplicate out. So double thanks!

PR #77, opened 2017-01-20
http->https redirector for web ui

nfriedly:
Solves this issue: https://forum.rockstor.com/t/automatic-http-https-redirection/2052

Tested by installing it locally and suddenly http://rockstor.local/ works :)

### DIFF
--- a/redirect-http-to-https.json
+++ b/redirect-http-to-https.json
@@ -1,0 +1,25 @@
+{
+    "HTTP to HTTPS redirect": {
+        "containers": {
+            "redirect-http-to-https": {
+                "image": "hope/redirect-http-to-https",
+                "launch_order": 1,
+                "ports": {
+                    "80": {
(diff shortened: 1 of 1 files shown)


magicalyak:
I really like this idea, as I prefer the redirect also and this gives you the option.  Ideally perhaps we will have a feature toggle in the options, but until then, this is a great concept.  Do we already have the MIT license in Rockstor?  If so we are fine (not sure why the maintainer used the MIT license).  The repo looks good, it's alpine running the reflector.  Can we change the repo to geldim/https-redirect (it has +1M pulls) and no worries on the license.  I tested the same json with that repo and it worked great.

nfriedly:
Switched ;)

magicalyak:
Just need to wait on @schakrava to merge at this point, looks good to me!

schakrava:
We'll add this as a feature in rockstor-core, but happy to merge this in the meantime. Thanks @nfriedly and @magicalyak

PR #79, opened 2017-01-24
Create crashplan.json

alazare619:
Crashplan is a local and cloud backup solution. It is managed via a program on a client computer. Rock-On has been tested and confirmed working flawlessly. Additional storage is allowed and recommended so it can backup all of the shares you give it.


magicalyak:
The json is missing the storage and TZDATA, there is another more recent pull request that addresses that (sorry for the delay in taking a look).

PR #80, opened 2017-01-24
Create ombi.json

alazare619:
Ombi is a webui request system for Plex. It supports all major automated PVR systems such as sonarr, radarr, couchpotato, headphones and sickbeard and friends. It has been tested and shows to be stable.


magicalyak:
Looks good but a few suggestions.  the --restart=always can be removed, it's default for rockstor so just remove.  Can you add the UID/GID lines in the ENV so you can specific a user to run as (I know I have a 1001:1001 plex user I use for all PVR stuff so I'd need that.)   And an optional sugestion, would DEV be better than EAP (it just is more obvious is why). Looks like you can do either.  Also sorry for the delay in anyone taking a look.

magicalyak:
I just realized DEV and EAP might be different!  If so, EAP makes more sense and I would not change.  I think the options for API token are overkill so I wouldn't add those (you haven't, just wanted to mention in case you were thinking).  I'd still add the PID/GID stuff and remove --restart (which is needed if run outside rockstor, but rockstor defaults to this for all rockons so it's not needed).

magicalyak:
One more note on that API rate limiting (I had to look!) it is an issue with the base logging here, I doubt it will matter so I wouldn't add it. https://github.com/tidusjar/Ombi/issues/581  again, not related but wanted to mention it in case someone is curious

magicalyak:
Last suggestion, can you move the description and version field to the bottom to be in line with other rockons?  This is an example - https://github.com/rockstor/rockon-registry/blob/master/nzbget.json
So along with the UID/GID and removing restart it looks good.

schakrava:
Thanks for your review @magicalyak!

@alazare619 really appreciate your contribution. I look forward to your updates to this PR and I'd be happy to merge soon as it's ready!

alazare619:
Ombi seems to work at this point however with every ombi docker I've tried on multiple systems it seems to randomly just not start. At this point I say we hold this as this file works well. But I dont feel ombi docker is stable enough at this point to consider it ready for mainstream.

magicalyak:
Thanks for explaining, I have two rockons in the same boat.  I'll see if I can figure out anything when I get some free time.

alazare619:
Looks like linuxserver.io has fixed the stability issue with ombi now @magicalyak

magicalyak:
Hoping to test today, I'm sorry for the delay.

alazare619:
@magicalyak  hey no problem brother :)

magicalyak:
Want to just add something in the info section? It says TBA, other than that it works fine for me!

magicalyak:
just need @schakrava to approve now, looks good to me.

alazare619:
@schakrava  can we get this merged in I have a few friends that keep pestering me wanting to know when it will be pushed.

schakrava:
Nice work @alazare619 and thanks for your contribution too, @magicalyak

PR #82, opened 2017-02-02
add zoneminder-1.30 rock-on

magicalyak:
This is for adding a zoneminder v1.30 rock-on. I couldn't get an in-place upgrade working due to changes with mysql (they use mariadb now).  I would also have to add a version select and a lot of if statements to get it to work.  Since people were asking for this, I wanted to get it out the door and that's why the version name is specified.  I had to use my own github repo, I've been looking for months for a solid one and couldn't find one.  I combined two people's work to get this to work.

### DIFF
--- a/zoneminder-1.30.json
+++ b/zoneminder-1.30.json
@@ -1,0 +1,45 @@
+{
+    "ZoneMinder-1.30": {
+        "containers": {
+            "zoneminder-1.30": {
+                "image": "magicalyak/docker-zoneminder",
+                "launch_order": 1,
+                "opts": [
+                         [
+                          "--privileged=true",
+                          ""
+                          ]
+                         ],
+                "ports": {
+                    "80": {
+                        "description": "Port for running ZoneMinder. You shouldn't open this externally (it's unsecure). Do not use 80 (it will interfere with RockStor)",
+                        "host_default": 8080,
+                        "protocol": "tcp",
+                        "label": "ZoneMinder Port",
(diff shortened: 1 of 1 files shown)


phillxnet:
@magicalyak Well done, this is great.

I agree that it's best not to over complicate it just to enable a dual version Rock-on. Do you think this is ready to be merged?

magicalyak:
I’d really like someone to try a camera with it, I don’t have a camera but it looks good otherwise.  The instructions should be the same for options (cam and ffmpeg under images) so that should be good.

Tom

phillxnet:
@magicalyak OK, yes no suitable camera here either, although a client is happily running your last rock-on for several months now, but can't really test there unfortunately. Maybe your forum post:
https://forum.rockstor.com/t/rock-on-repository-zoneminder-upgrade-to-1-30-from-1-29/2723/7
will get some takers / testers.

magicalyak:
This rockon should be good to go.  There is no reason it shouldn't work.  I don't have a camera to test but I didn't for the older zoneminder and it works.  I suggest it move forward since people have asked.  It is a seperate name/version due to the lack of a clear in-place upgrade.

schakrava:
Nicely done @magicalyak . Not a user myself, but installs fine and I really appreciate the instructions. detailed and well formatted. Thank you!

schakrava:
Forgot to mention, thanks for your review @phillxnet!

PR #83, opened 2017-02-05
Fix link for Transmission

gkadillak:


schakrava:
Thank you @gkadillak!

PR #86, opened 2017-02-11
New Rockons you may like

Jaco1960:
Transmission with OpenVPN and Crashplan with a WEB UI

### DIFF
--- a/crashplan-gfjardim.json
+++ b/crashplan-gfjardim.json
@@ -1,0 +1,58 @@
+ {
+	"Crashplan": {
+        "containers": {
+            "crashplan": {
+                "image": "gfjardim/crashplan",
+                "launch_order": 1,
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a share for the crashplan configuration & log files.",
+                        "label": "/config"
+                    },
+                    "/backup": {
+                        "description": "Choose a share for the crashplan backup files.",
+                        "label": "Backup"
+                    }
+				},
+				"opts": [ 
+					[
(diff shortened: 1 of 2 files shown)


magicalyak:
Can you explain a little more on the benefits of this?  In general we prefer multiple rockons so people can pick and choose items.  We also have a transmission and openvpn rockon that appears to work well.

Jaco1960:
The OpenVPN rockon is a OpenVPN server. The Transmission with OpenVPN rockon combines transmission with an OpenVPN client. This allows for downloading torrents via a VPN service so you do not reveal your IP address.

phillxnet:
@Jaco1960 Thanks for submitting this Rock-on and apologies for it not getting the attention it may have deserved.
As per the link made by @FroggyFlox it seems, for better or worse, that we now have this facility provided by way of @doenietzomoeilijk "Transmission with OpenVPN" Rock-on:
"Add Transmission through OpenVPN" #134
although from the looks of it your work here was an influence in that Rock-on.
Linking back to #133 as this (as per @FroggyFlox's opposite direction link) also shows this pr's reach.

Apologies but I am to close this pr as redundant even though it predates what ended up being merged.

Please re-open with additional comment if I have missed something obvious here.

Thanks again for you efforts, I'm afraid we have been found wanting in our maintenance of this repository and this pull request is likely a casualty.  Please don't let this put you off contributing to Rockstor in the future. All contributions are welcome and we are trying to step up our efforts re required maintenance of pull requests.

PR #87, opened 2017-02-19
New rock-on: Teamspeak3 server

Fredrik81:
New addition to rockon repository.
Teamspeak3 server is a popular VoIP application. There is one thing a bit tricky for new users with this application that the user needs to read logs in order to get admin access the server once it's installed.
Details on how to do that is added to the "more-info" part of the rock-on.

### DIFF
--- a/teamspeak3.json
+++ b/teamspeak3.json
@@ -10,21 +10,21 @@
                         "host_default": 9987,
                         "label": "Voice port",
                         "protocol": "udp",
-                        "ui": true
+                        "ui": false
                     },
                     "10011": {
                         "description": "Teamspeak3 Listning port (tcp). Suggested default: 10011",
                         "host_default": 10011,
                         "label": "Listning port",
                         "protocol": "tcp",
-                        "ui": true
+                        "ui": false
                     },
                     "30033": {
                         "description": "Teamspeak3 File-Transfer port (tcp). Suggested default: 30033",
                         "host_default": 30033,
                         "label": "File-Transfer port",
                         "protocol": "tcp",
-                        "ui": true
+                        "ui": false
                     }
                 },
                 "volumes": {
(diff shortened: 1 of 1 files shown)


magicalyak:
This looks really interesting, I'll try to find time to test soon.!

magicalyak:
I can't get this to work, I checked ports and noticed a few errors. I didn't open an issue with linuxserver.io yet but might later.  Everything runs with a complaint of not installing unzip because abc isn't sudo (which is weird because you can run everything without sudo as abc).  The permissions are right (I tried on two shares just to be sure).  I didn't try --net=host yet to check but I can't get this to work on my hardware.   Can you verify this works again (maybe on another box or have someone else check?)

magicalyak:
I added this to their issues - https://github.com/linuxserver/docker-gsm-ts3/issues/11

Fredrik81:
Hey thanks for testing it. I try it out again.
I tested using default settings in the rockon setup with a share dedicated to the TS3.
Also don't forget to change the share ACL rights before rockon setup.

The TS3 server started correct and i could connect. I have not tested to invite others to it but my own connection worked also i tested from my phone so it was accessable from the web.

After rockon setup there will be a long delay until the server is actually available. Not sure why i think it just take a long time to startup.

magicalyak:
How long? Maybe that's my problem...I'm just waiting for the web UI, I didn't get to the VoIP part.

Fredrik81:
Not sure the webinterface is included in the docker image. The button in the rockons interface will actually not do anything. I just don't think i can disable it. Have not seen any option to do that.

These are the ports that are open and as far as i know none of them are web-interface.
"9987" upd (Voice port)
"10011" tcp (Listning port)
"30033" tcp (File-Transfer port)

magicalyak:
Oh you should definitely disable the UI then.
On the ports, add the following  "ui": false
also remove this whole section
 "ui": {
             "slug": ""
         },

Fredrik81:
Make the changes but the button for webui is still shown in Rockstor. But now it will just take you to the IP of the rockstor server.

magicalyak:
Looks good, @schakrava just needs to merge. I don't have a voice system but the services work fine and the container runs well.

schakrava:
Thanks @Fredrik81 and @magicalyak , the Rock-on installs fine. I strongly recommend adding more help text in the settings. As a user who just installed this rock-on, I need to know what the next step is to start using the Rock-on? Perhaps a link to https://hub.docker.com/r/linuxserver/gsm-ts3/ in the info section or a summary of help listed there?

PR #88, opened 2017-02-21
Crashplan rockon

badriram:
Created a rockon for crashplan that works headless. It requires some active copying of credentials to a desktop with UI to make use of it. 
https://support.crashplan.com/Configuring/Using_CrashPlan_On_A_Headless_Computer

### DIFF
--- a/crashplan.json
+++ b/crashplan.json
@@ -38,7 +38,7 @@
             }
         },
         "description": "CrashPlan rockon, container from jrcs/crashplan",
-        "more_info": "<h4>Add more storage Crashplan to backup</h4><p>You can add more Shares to backup to crashplan rockon from the settings wizard of this Rock-on. Then, from crashplan UI on your <em>desktop</em> setup your backup. Refer to <a href="https://support.crashplan.com/Configuring/Using_CrashPlan_On_A_Headless_Computer">Crashplan: Using crashplan Headless</a></p>",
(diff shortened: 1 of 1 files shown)


magicalyak:
Since there is no ui, can you disable the slug section?  Otherwise this looks good and is from a well-used docker repo (250k pulls).

badriram:
Updated. thanks, was unsure if that section was required.

magicalyak:
Ok something's broken with the more_info section though.  It's the <a href section.  I'd take out the HTML markup and eliminate the " symbols.  It's basically closing the more_info on the JSON with the " right before https:// 
hope that makes sense.
So instead of
`<a href="https://support.crashplan.com/Configuring/Using_CrashPlan_On_A_Headless_Computer">Crashplan: Using crashplan Headless</a></p>",`
do
`https://support.crashplan.com/Configuring/Using_CrashPlan_On_A_Headless_Computer</a></p>",`
and it should work.

magicalyak:
Also don't need the TZ, just remove it.  However we already have a #79 pending so if that person modifies first, they might be merged instead (just a heads up).

magicalyak:
Oh otherwise great work, should have mentioned that.

badriram:
should be all set.

magicalyak:
Gotta wait for @schakrava to approve the merge but it looks good, thanks for working on it.

schakrava:
Thanks @badriram for your contribution and Thank you @magicalyak for your review!

PR #89, opened 2017-02-26
Fix broken URL for Transmission website

nfriedly:
there was an errant space in the middle of the link

### DIFF
--- a/transmission.json
+++ b/transmission.json
@@ -6,7 +6,7 @@
             "slug": ""
         },
         "version": "1.0",
-        "website": "http: //www.transmissionbt.com/",
+        "website": "http://www.transmissionbt.com/",
         "containers": {
             "transmission": {
                 "image": "dperson/transmission",


phillxnet:
@nfriedly Thanks for the pull request, and a nice find. But unfortunately we already have an outstanding pull request from @gkadillak which fixes the exact same issue.

I'm linking to the prior duplicate pull request here for context:
https://github.com/rockstor/rockon-registry/pull/83

nfriedly:
Op, sorry. Any reason why the other one hasn't been merged yet?

phillxnet:
@nfriedly No not really, merge master @schakrava tends to do merges in clumps. Especially in this repository. And from the looks of it another round of merges is due, so shouldn't be long now for a few of the pending pull requests to be addressed. Sorry for the inconvenience and thanks again for looking into and providing a sort for this.

PR #92, opened 2017-03-04
Add bitcoind rockon

eniLeg:
A simple rockon with the full node bitcoin server based on the [kylemanna/bitcoind](https://hub.docker.com/r/kylemanna/bitcoind/) docker image.

### DIFF
--- a/bitcoind.json
+++ b/bitcoind.json
@@ -1,0 +1,35 @@
+{
+    "Bitcoin": {
+        "containers": {
+            "bitcoind": {
+                "image": "kylemanna/bitcoind",
+                "launch_order": 1,
+                "ports": {
+                    "8333": {
+                        "description": "Listening port",
+                        "host_default": 28333,
+                        "label": "Port for incoming connections",
+                        "protocol": "tcp"
+                    },
+                    "8332": {
+                        "description": "JSONRPC port",
+                        "host_default": 28332,
+                        "label": "The JSONRPC server allows to query and control the server remotely",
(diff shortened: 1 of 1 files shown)


magicalyak:
Were you able to get the volume to persist with this? The README talks about creating and linking the volume container first but the rockon doesn't do that.  In other words, have you verified that destroying this container and recreating it leaves the blockchain data?

eniLeg:
I just tried these two cases:

- Switching the rockon off and on
- Switching off, uninstalling and installing the rockon again

In both cases the blockchain data + server configuration is persisted from the previous installation.

I am not sure if I did everything right in the rockon, so I would appreciate your input. This is how I installed and tested it:

- I created a separated share only for this purpose called `scratch`
- Installed the rockon and in the wizard I selected `scratch` as storage, I left the ports as they are in the defaults, 28333 and 28332.
- I created a port forwarding rule in the router from port 8333 to <local_rockstore_ip>:28333. 

On my side everything seems to work as expected, if I inspect the docker log I can see the live updates from bitcoind and the node appears as accessible from https://bitnodes.21.co.

magicalyak:
Can you add the requirements in the json.  You also might want to add the 100GB minimum to the /bitcoin volume.  The hardware reqs are pretty strong and we want to ensure people don't accidentally use this and not realize.

schakrava:
Thanks @eniLeg , I really want to give this a try. And thanks @magicalyak for your review!

PR #99, opened 2017-03-21
Zabbix xxl

magicalyak:
Zabbix XXL sets up a full Zabbix Server and MariaDB using three containers linked.  I was pretty impressed this worked and wanted to share.  It should be noted the DB container uses 1 GB memory though (but that's listed in the readme on the docker page). The password has to be hardcoded as we can't grab variables and hardcode at the same time.  Attempting to pass them did not work so I generated a hash to make the hardcoded DB password somewhat complex.  Considering the database is not using a mapped port and restricted to localhost, this should be secure enough to use I hope.

### DIFF
--- a/zabbix-xxl.json
+++ b/zabbix-xxl.json
@@ -15,6 +15,12 @@
                 "image": "busybox",
                 "label": "latest",
                 "launch_order": 1,
+                "opts": [
+		            [
+			            "--restart",
+			            "no"
+		            ]
+		        ],
                 "volumes": {
                     "/var/lib/mysql": {
                         "description": "Choose a Share for Zabbix Database. (create a share called zabbix-db for this purpose)",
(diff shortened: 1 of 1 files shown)


magicalyak:
This works great at this point!  Ready to merge.

zabbix-xxl.json
```diff
@@ -0,0 +1,123 @@
...
+                    [
+                        "-e",
+                        "MARIADB_PASS=766a2bb5d24bddae466c572bcabca3ee"
+                    ],
+                    [
+                        "--volumes-from",
+                        "zabbix-db-storage"
+                    ],
+                    [
+                        "-v",
```
schakrava: We do this automatically for all containers. Can you confirm it works without this -v option?

magicalyak:
@schakrava removed lines and verified it works.  It works.  The one option to not restart on the busybox is needed and works because it just runs once and quits for the db vol mapping.  It's the method used by the author and I didn't want to have to rewrite this just for that.

schakrava:
Thank you @magicalyak

PR #101, opened 2017-03-22
add lazylibrarian rock-on

coleberhorst:
added the lazylibrarian docker package from [https://hub.docker.com/r/linuxserver/lazylibrarian/linuxserver/lazylibrarian](url)

Modified headphones [https://github.com/rockstor/rockon-registry/blob/master/headphones.json](url) rock-on file since they are basically the same codebase.
Tested on windows uploading json via winscp to current stable version of rockstor. Worked well. No issues.

### DIFF
--- a/lazylibrarian.json
+++ b/lazylibrarian.json
@@ -1,0 +1,53 @@
+{
+  "LazyLibrarian": {
+    "containers": {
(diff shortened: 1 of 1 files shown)


schakrava:
Thank you @coleberhorst. App works well!

PR #103, opened 2017-04-01
Update Subsonic package name

JasonTarka:
`hurricane/docker-subsonic` doesn't exist on docker.io, but `hurricane/subsonic` does.
Likely an outdated reference.


schakrava:
Thank you @JasonTarka !

PR #107, opened 2017-05-08
Readme corrections and clarifications

nfriedly:
Removed "optional" from the descriptions of a couple of fields that are not actually optional and also clarified the options array field.


schakrava:
greatly appreciated @nfriedly !

PR #110, opened 2017-05-18
Utorrent - added changeable UID and GID from base docker spec

coleberhorst:
Had some issues with utorrent running as default 666 UID and GID. The base docker has environment variables for  UID and GID, which I added into the json.


schakrava:
Thanks @coleberhorst

PR #112, opened 2017-05-21
adding more optional descriptions to json spec

coleberhorst:
adds specifications for optional ui and slug object at top level and ui value in port object.


schakrava:
Thank you @coleberhorst

PR #115, opened 2017-06-12
changed radarr spec to allow for storage volume changes. 

coleberhorst:
Needed to run custom scripts not located in default directories for docker and just to allow for media to be placed in multiple directories.


schakrava:
LGTM. Thank you @coleberhorst !

PR #118, opened 2017-07-18
add resilio sync rock-on

### LINKED ISSUE #117: btsync rockon should be replace by resilio/sync
bittorrent/sync docker image has been deprecated and replaced by resilio/sync
https://hub.docker.com/r/resilio/sync/

Deprecated image stops at version 2.3.8 which causes licences issues if all your other peers are on 2.4 using latest Resilio licence model.

Can we expect an upgrade soon? Does changing the Rockon's name and image will cause upgrade problems?

Thanks

jybleau:
To fix #117 

Tested locally, working fine.

It uses the official Docker container: https://hub.docker.com/r/resilio/sync/

To update an existing btsync rock-on installation: 

- uninstall btsync rock-on
- install resilio sync rock-on
- use the same configuration as for your previous btsync installation (ports and share(s)).

### DIFF
--- a/resilioSync.json
+++ b/resilioSync.json
@@ -1,0 +1,39 @@
+{
+  "Resilio Sync": {
+    "containers": {
+      "resilio-sync": {
+        "image": "resilio/sync",
+        "launch_order": 1,
+        "ports": {
+          "55555": {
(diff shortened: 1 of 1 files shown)


schakrava:
Thanks @jybleau and @magicalyak

PR #120, opened 2017-07-24
Adding in minio.io storage provider.

jmhardison:
Minio.io is a distributed object storage server providing S3 compatibility.
I was not able to find where I can enter a launch command for docker run outside of what I saw in the core code that had custom extends for specific images. So I've created a custom docker image that sources from minio/minio and changes the launch to server mode. (Build hooks set to minio/minio so it will always remain up to date with core.)

This container will expose 9000/tcp and use several environment variables to turn on the portal and set user/pass. (The user and password get generated automatically if not defined, which would be in the console log of the container and not exposed.)

Check http://minio.io for more info.

Testing simply can be performed by enabling the web ui, and then interacting with the portal. Otherwise you can follow examples for console or code from minio.io.

### DIFF
--- a/minio.json
+++ b/minio.json
@@ -1,0 +1,52 @@
+{
+    "Minio": {
+        "containers": {
+            "Minio": {
+                "image": "jhardison/rockstor-minio",
+                "launch_order": 1,
+                "ports": {
+                    "9000": {
+                        "description": "Port for Minio services including API and Portal access. Suggested default: 9000",
+                        "host_default": 9000,
+                        "label": "Minio main port",
+                        "protocol": "tcp",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/root/.minio": {
+                        "description": "Minio configuration share.",
+                        "label": "Config Storage",
+                        "min_size": 1073741824
+                    },
+                    "/export": {
+                        "description": "Choose a share for all data in Minio.",
+                        "label": "Data Storage"
+                    }
+                },
+                "environment":{
+                    "MINIO_ACCESS_KEY":{
+                        "description": "Custom username or access key of 5 to 20 characters in lenght.",
+                        "label": "Custom Access Key"
+                    },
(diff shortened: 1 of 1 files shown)


magicalyak:
Any reason the base docker minio couldn't be used?  We try to stay away from any docker hub custom builds when we can, generally look for project owner, linuxserver.io first if possible.

jmhardison:
If I remember right it was because the base image does not have defaults for the mode (server) and export path, and from what I could tell a rockon didn't have the ability to do command and arguments that follow a docker run command. (currently) Be happy to revert to the root for that, but those fields are required to launch correctly. Unless I completely missed where/how that can be done in a rockon. I saw of course the ability to do options such as environment, but not command arguments.

Linuxserver.io also does not have a drop for this. I could approach minio and see if they are willing to implement defaults when no mode or argument is provided to the executing container to default it to server mode with a path. (Effectively mirror their example command into the default state given a run with no args.)

Feel free to sits as I chase that, if they accept I can either report back or update the request back to root image once merged.

jmhardison:
Well that's a more direct no.
https://github.com/minio/minio/pull/3030

> We had to make this breaking change since minio has other sub-commands, it was our mistake that "server" was included as part of the entrypoint earlier.
To run docker properly now the right command is
docker run -p 9000:9000 minio/minio server /export

and

> harshavardhana commented on Oct 21 2016 
Thanks for the patch @olalonde - but we are not going to default any commands for docker container . Docs become inconsistent with regular binary releases

I'll go submit a request to linuxserver.io and see if they want to pull it into themselves. End of the day, either that or if there is a way to make rockons use custom args in a launch command.

magicalyak:
I think that's adequate explanation. I'll review soon using your container.

jmhardison:
As work is being completed to add argument support, I'm going to close out this pull request.

FroggyFlox:
Glad the argument support could be of some help to you, @jmhardison. Please let me know if it is still lacking in any way.

PR #124, opened 2017-08-14
qBittorrent

absentbri:
add qBittorrent


phillxnet:
@absentbri Thanks for the Rock-on submission. I have recently gained merge capabilities on the rock-on repository and would like to merge this pr. However we first need a review.

Could you or anyone else familiar with this software confirm the validity of this pull request?

absentbri:
As far as I know this is still good to go. I moved to UnRaid about a year ago so I've no way of confirming. This was in 2017...

phillxnet:
@absentbri As per you other pr, thanks for confirming and for your efforts.

Requesting a review of this pr so we can get it merged. Marking a 'fix' for our backlog pr maintenance.

absentbri:
I have made the changes @FroggyFlox advised, I had already fixed this in the system I am using these days but this was so long ago now... Hoping this being available helps someone so figured making the edits was worth it.

phillxnet:
@absentbri Thanks for making these changes and thanks to @FroggyFlox for the review / recomendations.
I've now tried this Rock-on via local repo (/opt/rockstor/rockons-metastore) and it appears to work a treat so that great. Also turns on and off as expected.

But we have one last usability addition text wise. Given the requirement for port selection of 8090 (to align with the currently hard wired "-e WEBUI_PORT=8090" element, we should indicate this in that ports description section i.e. adding  " try 8090 here if unsure." Or whatever better wording you come up with. Bit messy but should work for now. We need to enhance our Rock-on system to be more flexible: which is actually an ongoing mission of @FroggyFlox .

With this addition I think we can pop this one in. Let me know if you are game to make these changes or I can edit via another commit.

Thanks again for tending to this, it looks like a nice addition Rock-on wise, and thanks for adding @FroggyFlox's earlier suggestions. 
Nice touch moving to 8090 by the way. 

I used: the following setting when testing:

qbittorrent

FroggyFlox:
@phillxnet, thanks for testing it.

I'm out of time right now but beware the port choice. In my testing VM, for instance, 8090 was used by something else, which caused the "white page" issue. Setting the -e WEBUI_PORT to the same is not enough in this case and both host and container ports need to be changed in this case.
Maybe we need to set a port different than 8090 as default container port (one unlikely to be used).
I'll get back later with more details, sorry for the rush.

FroggyFlox:
Back with a little more details and testing.

The issue described in my previous comment refers to the following line in the [image's description](https://hub.docker.com/r/linuxserver/qbittorrent/) on docker Hub: 

> Due to issues with CSRF and port mapping, should you require to alter the port for the webui you need to change both sides of the -p 8080 switch AND set the WEBUI_PORT variable to the new port. 

In my case, port 8080 was currently used so Rockstor proposed a default of 8081 and I encountered the "white page" error described above. Unfortunately, simply setting a new `"environment"` object in the JSON for `WEBUI_PORT` isn't sufficient as the value of this variable **as well as** the container **AND** host ports for the webUI must match.
Unfortunately, we are not flexible (yet, as this is on now on my to-do list) on the host-port setting, which means that the only workaround I can think of for now is to set the webUI port to an "exotic" number in an attempt to limit to potential for having it already used in _most_ users (if not _all_ users). We could set all of this "in the background" but as need to set the webUI port to be used for "ui" (`"ui": true), it needs to be part of the `"ports"` object and not `"opts"`.
I'll submit a working suggestion as an example.

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                "image": "linuxserver/qbittorrent",
+                "launch_order": 1,
+                "opts": [
+                    [
+                        "-e",
+                        "UMASK_SET=007"
+                    ],
+                    [
+                        "-e",
+                        "WEBUI_PORT=8090"
```
FroggyFlox: Set WEBUI_Port to an exotic number matching the published `"ui"` port below.
```suggestion
                        "WEBUI_PORT=12500"
```

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                    ]
+                ],
+                "ports": {
+                    "8090": {
```
FroggyFlox: Set webUI port to an exotic number **hopefully** unused on most users' machine:

```suggestion
                    "12500": {
```

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                        "description": "qBittorrent WebUI port. When changing this you must also change environment \"WEBUI_PORT\" setting.",
+                        "host_default": 8090,
```
FroggyFlox: Must match container port set above.
```suggestion
                        "host_default": 12500,
```

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                        "UMASK_SET=007"
+                    ],
+                    [
+                        "-e",
+                        "WEBUI_PORT=8090"
+                    ]
+                ],
+                "ports": {
+                    "8090": {
+                        "description": "qBittorrent WebUI port. When changing this you must also change environment \"WEBUI_PORT\" setting.",
```
FroggyFlox:
```suggestion
                        "description": "qBittorrent WebUI port. MUST be 12500.",
```

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                        "host_default": 8090,
+                        "label": "qBittorrent WebUI port",
```
FroggyFlox: Shorten label:
```suggestion
                        "label": "WebUI port",
```

qbittorrent.json
```diff
@@ -0,0 +1,64 @@
...
+                        "protocol": "tcp",
+                        "ui": true
+                    },
+                    "6881": {
+                        "description": "Incoming data (DHT). You may need to open it(protocol: udp) on your firewall. Suggested default: 6881. Should NOT be changed",
+                        "host_default": 6881,
+                        "label": "qBittorrent DHT port"
```
FroggyFlox: Shorten label:
```suggestion
                        "label": "DHT port"
```

phillxnet:
@absentbri Apologies to drag this one up again but if you are game? @FroggyFlox earlier this year, gave a few suggestion on how we could finally get this on in. If you fancy committing those suggestions ("Commit suggestions" button) we can merge. We like the linuxserver.io images and it's still maintained so it would be nice to not have wasted your's and @FroggyFlox's efforts on this one. Assuming @FroggyFlox you are still happy with your suggestions.

I know their are exceptions in this Rock-on but if they are stated, as per @FroggyFlox more recent suggestions, it could work for general use. Which would be good.

Really trying to get our Rock-ons backlog down as it's part of the problem, that and a million and one distractions in a project of this size of course.

absentbri:
While I don't feel setting the port as requested I'll never be using it so accepted it all. There will always be a potential for conflict with Docker containers so leaving it stock and offering a note on how to change it was probably more advisable. 

At any rate I approved all the changes as I'm no longer as part of this project.

PR #127, opened 2017-09-14
moved syncthing to linuxserver version

magicalyak:
Syncthing is now updated to use the linuxserver.io version and can specify UID and GID.  The mapping of the /config share is perfectly compatible with the older mapping of the previous maintainer.  In other words, you can use this json with the older version and keep all settings and not lose anything.

### DIFF
--- a/syncthing.json
+++ b/syncthing.json
@@ -2,7 +2,7 @@
     "Syncthing": {
         "containers": {
             "syncthing": {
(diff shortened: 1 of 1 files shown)


alazare619:
great commit linuxserver is a great team and produce great docker containers.

schakrava:
Thanks @magicalyak !

PR #128, opened 2017-09-20
Add portainer.json

alazare619:
Portainer is an advanced docker cluster,swarm, and single node management framework. Its very advanced and extremely lightweight. It integrates perfectly into rockstor i've been using it for a few weeks now without any issue managing the advanced tasks for my rockstor docker node.

### DIFF
--- a/booksonic.json
+++ b/booksonic.json
@@ -1,0 +1,58 @@
+{
+    "booksonic": {
+        "containers": {
+            "booksonic": {
+                "image": "linuxserver/booksonic",
+                "launch_order": 1,
+                "ports": {
+                    "4040": {
+                        "description": "WebUI port. Suggested default: 4040",
+                        "host_default": 4040,
+                        "label": "WebUI port",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a Share for configuration. Eg: create a Share called booksonic-config for this purpose alone.",
+                        "label": "Config Storage"
+                    },
+		    "/books": {
+			"description": "Choose a Share for books. Eg: create a Share called booksonic-books for this purpose alone.",
+			"label": "Books Storage"
+		    },
+
+		    "/podcasts": {
+			"description": "Choose a Share for podcasts. Eg: create a Share called booksonic-podcasts for this purpose alone.",
+			"label": "Podcasts Storage"
+                    },
+
+		    "/media": {
+			"description": "Choose a Share for media. Eg: create a Share called booksonic-media for this purpose alone.",
+			"label": "Media Storage"
+		    }
+		},
+                "environment": {
+                    "PUID": {
+                        "description": "Enter a valid UID to run booksonic with. It must have full permissions to all Shares mapped in the previous step.",
+                        "label": "UID",
+                        "index": 1
+                    },
+                    "PGID": {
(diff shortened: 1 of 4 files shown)


FroggyFlox:
How does the volume creation/management function on top of Rockstor's shares? I've been tempted and a little interested by Portainer a few weeks ago but wasn't sure how to deal with this part... Would it better to create volume from within Portainer or simply bind to host's share previously created from Rockstor? I would tend to favor the latter in order to keep snapshot possibility, etc..., but I was curious about your experience and use with this so far.

magicalyak:
You're likely going to need to add the --privileged also.  Change the opts to:

```
                "opts": [
                    [
                        "-v",
			"/var/run/docker.sock:/var/run/docker.sock"
                    ],
		    [
			"--privileged",
			""
		    ]
                ],
```

magicalyak:
And I'd add a a volumes map also:

```
                "volumes": {
                    "/data": {
                        "description": "Data share for Portainer information.",
                        "label": "Data Storage"
		    }
		},
```

alazare619:
Weird I didnt get any emails regarding this... Anyways so far I haven't had to have anything else out of the normal. as far as volume maps go yes we could add extra volumes but all the data is actually getting stored in the sock file based on how portainer manages their container...its weird... I haven't needed the privledged yet but I will add that as a CYA

alazare619:
@FroggyFlox  to answer your question I created portainer after I created a bunch of rock-ons already and it managed the existing ones fine. I've created them within portainer as well and they are fine. So either way is fine.

alazare619:
@magicalyak Added the persistent data volume and added it as privileged, I've not had any issues with SELINUX and it yet however without but added it just incase.

@FroggyFlox to answer your question about volumes. I've found it easier to start the dockers by hand via the cli it can use rockstors shares there. Management after the initial creation I do via portainer. I've for some time outgrew the functionality of rockstors management for docker and this definetly supplements it quiet a bit. To clarify I create the shares via rockstor. Then do the initial creation of the docker container via the cli. After that I manage all aspects of the docker via portainer.

portainer.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "Portainer is a lightweight advanced management UI which allows you to easily manage your Docker host or Swarm cluster.",
+    "containers": {
+      "portainer": {
+        "image": "portainer/portainer",
+        "launch_order": 1,
+        "opts": [
+          [
+            "-v",
+            "/var/run/docker.sock:/var/run/docker.sock"
+          ][
```
doenietzomoeilijk: There's a comma missing here, making the file invalid JSON.

portainer.json
```diff
@@ -0,0 +1,37 @@
...
+            ""
+          ]
+        ],
+        "ports": {
+          "9000": {
+            "description": "Portainer WebUI port. Suggested default: 8090",
+            "host_default": 9000,
+            "label": "WebUI port",
+            "protocol": "tcp",
+            "ui": true,
```
doenietzomoeilijk: There's a trailing comma here, making the file invalid JSON.

alazare619:
@schakrava

alazare619:
@phillxnet  can you test this and comit if it all tests well?

alazare619:
closing as this seems to of died.

PR #129, opened 2017-09-29
Subsonic Update for GID and UID

magicalyak:
This is replacing the image for subsonic with one that handled GID and UID.


magicalyak:
This needs more testing, don't merge yet.

PR #130, opened 2017-09-29
add booksonic rockon

magicalyak:
Added BookSonic to the rockon registry, it includes GID/UID and is a book streaming server.  used the linuxserver container


schakrava:
Thanks @magicalyak !

PR #131, opened 2017-10-17
Add gmvault rockon configuration

epowell:
This is a working rockon configuration for running Gmvault, a GMail backup utility. See: [gmvault.org](http://gmvault.org/).

It uses a slim Docker image by the utility's creator,  @guillaumeaubert : [aubertg/gmvault-docker](https://hub.docker.com/r/aubertg/gmvault-docker/). The included instructions there are a good accompaniment to the rockon install process.

Initial setup does require some work. Specifically, you have to run gmvault manually on a machine with a browser in order to generate the initial OAuth token, then store that token file in the root of the /data share.

### DIFF
--- a/gmvault.json
+++ b/gmvault.json
@@ -1,0 +1,78 @@
+{
+    "Gmvault": {
+        "containers": {
+            "gmvault": {
+                "image": "aubertg/gmvault-docker",
+                "launch_order": 1,
+                "volumes": {
+                    "/data": {
+                        "description": "A share where the OAuth token file exists, and email backups and logs will be stored.",
+                        "label": "Data Storage"
+                    },
+                    "/etc/ssmtp/": {
+                        "description": "A directory containing only a ssmtp config file to send email reports from the Docker container.",
+                        "label": "SSMTP Configuration"
+                    }
+                },
+                "environment": {
+                    "GMVAULT_EMAIL_ADDRESS": {
+                        "description": "The email address of the account to back up.",
+                        "label": "GMVAULT_EMAIL_ADDRESS",
+                        "index": 1
+                    },
+                    "GMVAULT_SEND_REPORTS_TO": {
+                        "description": "The email address to send reports to; defaults to GMVAULT_EMAIL_ADDRESS.",
+                        "label": "GMVAULT_SEND_REPORTS_TO",
+                        "index": 2
+                    },
+                    "GMVAULT_UID": {
+                        "description": "Numeric uid in the host that should own created files; defaults to 9000.",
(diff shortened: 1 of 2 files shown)


phillxnet:
@epowell Thanks for submitting this: nice. One concern I have is the current activity of the project, ie from their GitHub page it looks like the last commit was 18 month ago:
https://github.com/gaubert/gmvault/commits/master
however their 1.9.2 branch last commit is more like 7 month ago:
https://github.com/gaubert/gmvault/commits/gmv-1.9.2

Is this still dependable code.

The container itself seems to install gmvault from PyPI via "pip install gmvault":
https://pypi.python.org/pypi/gmvault/1.9.1

Might it be possible to have a link or short pointer in the info section on initial setup requirements, ie how the OAuth token can be generated and put into effect in a Rock-on context. I.e. something akin to the instructions laid out in the zoneminder Rockon:
https://github.com/rockstor/rockon-registry/blob/master/zoneminder.json

This would be a great Rock-on to get in so thanks for your submission and let us know your opinion on the questions above and if you fancy adding the more explicit OAuth instructions. We are after making Rock-on installs as painless as possible: ideally.

epowell:
Absolutely, the more_info block is a good place for the setup documentation. Thanks for the pointer to zoneminder.

Regarding activity, I don't know if it will be updated much or not. But it does work fine as-is for most cases. I haven't seen any issues on my account with 2FA and several gigs of email. Other tools for this have been out of date since ~2012, so this is relatively recent for the space I'd say. I'll leave the final determination up to you, but I am satisfied with the results I have seen locally.

More info (literally, more_info) coming soon!

epowell:
No worries about the delay, this fell off my radar after I intended a documentation update. So my apologies as well!

I've recently lost my original OAUTH token and need to spend a bit of time updating it, which would be a good opportunity to update the doc. If you'd like to wait for that - knowing that I may not get to it for quite a while - it's fine by me, or if you're happy with the current state that is also fine.

phillxnet:
Closing pr as the project is now several years old without any commits and has privileged access to folks google account.

Please re-open if you think we can resurrect this one. It's a really nice fit to Rockstor as well but never mind.

@epowell My apologies for failing you on this one. Do please consider submitting a new pr if you can find an actively supported image/software for this. Again my apologies for dropping the ball on this one.

epowell:
@phillxnet No worries at all! I also dropped the ball on the docs here, and soon afterward the GMvault software stopped working for me, and wasn't important enough for me to re-engage. So I think you allocated the right amount of effort here. ;) 

Appreciate you cleaning up the PR.

PR #133, opened 2017-10-22
Add Transmission + OpenVPN Rock-on.

doenietzomoeilijk:
I wanted to be able to use Transmission through an OpenVPN tunnel, and implementing [haugene/docker-transmission-openvpn](haugene/docker-transmission-openvpn) as a Rock-on seemed like the easiest way to do it.

To keep things flexible, I've added a "custom" OpenVPN profile share by default, allowing you to add your own `.ovpn` files (and the certs and keys that come with it, if you don't inline it) as outlines [here](https://github.com/haugene/docker-transmission-openvpn#using-a-custom-provider).

I've only tested it with the VPN provider I have (VPN-Secure); I assume it works with other providers as well (as well as the original Docker container would), but I'd like some folks with VPNs to test that, if possible.

### DIFF
--- a/root.json
+++ b/root.json
@@ -46,6 +46,7 @@
     "TeamSpeak3": "teamspeak3.json",
     "TFTP server": "tftpserver.json",
     "Transmission": "transmission.json",
+    "Transmission + OpenVPN client": "transmission-openvpn.json",
     "Unifi Controller": "unifi.json",
     "Xeoma Video Surveillance": "xeoma.json",
     "ZoneMinder": "zoneminder.json",
--- a/transmission-openvpn.json
+++ b/transmission-openvpn.json
@@ -1,0 +1,93 @@
+{
+    "Transmission - OpenVPN": {
+        "website": "https://hub.docker.com/r/haugene/transmission-openvpn/",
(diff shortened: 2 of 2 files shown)


FroggyFlox:
This seems very similar to commit d29837a5307df3f5c713d80b8a8add51f02c2610 in the PR #86. It is by itself, however, and seems to be with a separate .ovpn conf file, which seems to be an advantage over the other one. Am I correct?

doenietzomoeilijk:
@FroggyFlox Correct, this is "just" one Rock-on with one goal, and with support for a custom provider, which was my primary use case. @Jaco1960 did however add some more environment values to control Transmission with, which I might want to add to this PR as well, although the Transmission config lives on a share and is editable.

doenietzomoeilijk:
It seems like the container doesn't really work after a reboot of my Rockstor server; I'll investigate and see whether it's a tunnel thing or some other issue. I might add the (provided) reverse proxy for the image after all, to make this easier and do away with the LOCAL_NETWORK setting.

Do not merge until further notice (If you'd prefer me to close the PR and reopen it later, instead, just let me know).

PR #134, opened 2017-10-31
Add Transmission through OpenVPN

doenietzomoeilijk:
Re-opening #133 but this time from a branch. I'm working on getting the proxied version going as well, but in the meantime, this *is* a viable solution. 

For futher description and previous discussion, see #133 .


schakrava:
Thank yoiu @doenietzomoeilijk

PR #138, opened 2017-12-19
Duckdns

alazare619:
Duckdns is a Dynamic DNS service. This container has no storage volumes necessary everything is stored in enviromental variables. The service is 100% free..

### DIFF
--- a/duckdns.json
+++ b/duckdns.json
@@ -1,0 +1,35 @@
+{
+	"DuckDNS": {
+		"website": "https://hub.docker.com/r/linuxserver/duckdns/",
+		"version": "1.0",
+        "description": "Duck DNS is a free service which will point a DNS (sub domains of duckdns.org) to an IP of your choice. The service is completely free, and doesn't require reactivation or forum posts to maintain its existence.",
+		"containers": {
+			"duckdns": {
+				"image": "linuxserver/duckdns",
+				"launch_order": 1,
+				"environment": {
+					"PUID": {
+						"description": "Enter a valid UID to run DuckDNS as.",
+						"label": "UID to run Duck DNS as.",
+						"index": 1
+					},
+					"PGID": {
+						"description": "Enter a valid GID to use along with the above UID. ",
+						"label": "GID to run Duck DNS as.",
+						"index": 2
+					},
+					"SUBDOMAINS": {
+						"description": "Enter a valid Subdomain. Multiple subdomains allowed, comma separated, no spaces",
+						"label": "Subdomain(s)",
+						"index": 3
+					},
+					"TOKEN": {
+						"description": "Authentication/API Token for DuckDNS",
+						"label": "Token for Duck DNS",
+						"index": 4
+				    }
+			    }
+		    }
+	    }
+    }
+}


FroggyFlox:
Tested this on my Rockstor instance and works perfectly.
I did submit a pull request on @alazare619 branch to add the possibility for the user to specify its timezone for correct timestamps reporting in the logs.

phillxnet:
@alazare619  and @FroggyFlox so what's the intention with the Timezone addition, it seems to have been added and then removed?
The rock-on / pr itself looks to be a nice addition.

alazare619:
I added it. Then remembered it uses the rockstor local time by default. It seems counter intuitive to add a required enviromental variable when your system timezone should be set correctly anyways.

phillxnet:
@alazare619 Are yes, good point. Forgot about that, thanks.

florianduclot:
@phillxnet My bad... that was my fault... I tested this Rock-on on my test rockstor instance in which the timezone is incorrect (I suffered from an issue I saw someone having on the forum a while back in which the install proceeded automatically without giving me the chance to change anything). As a result, I was seeing "wrong" timezones in my DuckDNS logs, which prompted me to submit a PR to @alazare619 's JSON. It turned out to be un-necessary so removing it kept the rock-on more simple.

florianduclot:
The only "complaint" I would have with this Rock-on is that the IP is updated every 5 mins--a setting that cannot be changed at first glance. I would thus assume this will force the disk with the `rockon-root` to be active at least every 5 min and thus never go idle... Am I correct?
If it is indeed the case, this may warrant an appropriate note in the rock-on "additional information" section... thoughts?

Otherwise the rock-on is working perfectly on my test system.

alazare619:
you could potentially change the update frequency but linuxserver.io has it hard coded. You would need to docker exec /bin/bash duckdns then go in and edit the scheduled task via crontab.

ref crontab job: https://github.com/linuxserver/docker-duckdns/blob/master/root/defaults/duckcron

FroggyFlox:
@phillxnet My bad... that was my fault... I tested this Rock-on on my test rockstor instance in which the timezone is incorrect (I suffered from an issue I saw someone having on the forum a while back in which the install proceeded automatically without giving me the chance to change anything). As a result, I was seeing "wrong" timezones in my DuckDNS logs, which prompted me to submit a PR to @alazare619 's JSON. It turned out to be un-necessary so removing it kept the rock-on more simple.

FroggyFlox:
The only "complaint" I would have with this Rock-on is that the IP is updated every 5 mins--a setting that cannot be changed at first glance. I would thus assume this will force the disk with the rockon-root to be active at least every 5 min and thus never go idle... Am I correct?
If it is indeed the case, this may warrant an appropriate note in the rock-on "additional information" section... thoughts?

Otherwise the rock-on is working perfectly on my test system.

EDIT: repost due to account mixup... sorry.

FroggyFlox:
@alazare619 thanks for the reference.

I noticed the second ranked DuckDNS image has an interval variable: https://hub.docker.com/r/meltwater/duckdns/

Do you think it's worth using that one instead or not worth the hassle? Sorry for all the questions, but I'm unfamiliar with what disk in Rockstor would be active when the no volumes are bound in a container.

alazare619:
since theres actually no config or anything for logs to be stored its actually just writing to ram so really there should be 0 io on any of the disks. I've not ran an iotop to verify but im sure it would confirm my thoughts that this rockon shouldn't write to any disk at all and should run entirelly in ramdrive (reserved ram)

FroggyFlox:
> it should be /var/log/daemon

So only the `rockstor_rockstor` pool then, isn't it? If I understand correctly this means it shouldn't have any effect on idling disk(s) other than the OS disk.

Thanks a lot for your time and effort @alazare619 !

alazare619:
exactly

FroggyFlox:
I've been running this Rockon on my main Rockstor instance (running 3.9.2-32) and works like a charm.

phillxnet:
Thanks @alazare619 much appreciated.  Apologies for the unreasonable delay in merging this.
And thanks @FroggyFlox for the review / testing.
Merging main json definition file now and will add root.json component shortly as part of #171

PR #139, opened 2017-12-19
Smokeping

alazare619:
Smokeping is a network monitor and history graph.  You can setup multiple sites for it to ping. It will log the latency and graph it for historical purposes. This is a must for anyone having issues with random latency or ISP issues. This container can NOT be ran as root.

Note: "slug": "smokeping/smokeping.cgi" does not seem to work the FQURL should be host:port/smokeping/smokeping.cgi

### DIFF
--- a/smokeping.json
+++ b/smokeping.json
@@ -2,7 +2,10 @@
 	"SmokePing": {
 		"website": "https://hub.docker.com/r/linuxserver/smokeping/",
 		"version": "1.0",
-        "description": "SmokePing is a network latency history monitor.",
+		"description": "SmokePing is a network latency history monitor.",
+		"ui": {
+			"slug": "smokeping/smokeping.cgi"
+		},
 		"containers": {
 			"SmokePing": {
 				"image": "linuxserver/smokeping",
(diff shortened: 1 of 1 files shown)


schakrava:
The ui slug is in the wrong section. Please see ghost.json for reference

alazare619:
I'll test this out tonight once i'm home and report back

smokeping.json
```diff
@@ -0,0 +1,48 @@
...
+				"image": "linuxserver/smokeping",
+				"launch_order": 1,
+				"volume_add_support": true,
+				"ports": {
+					"80": {
+						"description": "SmokePing WebUI port. Suggested default: 7878",
+						"host_default": 80,
+						"label": "WebUI port",
+						"protocol": "tcp"
+					}
```
FroggyFlox: The ui_slug is correct, but you need to set this port as the webUI port in order for Rockstor to create the appropriate "webUI" button.
The `"ports"` section simply needs an extra line with `"ui": true`, as follows:
```
"protocol": "tcp",
"ui": true
```

phillxnet:
@FroggyFlox Thanks for reviewing this promising Rock-on. Hopefully @alazare619 can step in and make the required change, would be nice to see this to fruition.

phillxnet:
@FroggyFlox and @alazare619 In the interests of further clearing up our backlog I'm going to merge this as is and apply @FroggyFlox suggested fix. Then add it's root.json entry. That way we can get it done and dusted.

PR #141, opened 2017-12-19
add transcode volume option to match LSIO spec

alazare619:
the current docker container spec from linuxserver.io has an option for transcode volume. Update the JSON to match the spec.

### DIFF
--- a/plex.json
+++ b/plex.json
@@ -26,6 +26,10 @@
                     "/data": {
                         "description": "Choose a Share for Plex content(your media). Eg: create a Share called plex-data for this purpose alone. You can also assign other media Shares on the system after installation.",
                         "label": "Data Storage"
+                    },
+                    "/transcode": {
+                        "description": "Choose a Share for Plex transcoding content. Eg: create a Share called transcode the faster the drive the better for this purpose.",
+                        "label": "Transcoding Storage"
                     }
                 },
                 "environment": {


phillxnet:
@alazare619 Thanks for recent flurry of rock-on enhancement but are you sure this is still implemented as from the changelog at:
https://hub.docker.com/r/linuxserver/plex/
we have:
```
07.04.16: removed /transcode volume support (upstream Plex change) and modified PlexPass download method to prevent unauthorised usage of paid PMS
24.09.15: added optional support for volume transcoding (/transcode), and various typo fixes.
```
I.e it was in there for a bit and then they removed it as upstream removed it by the looks of it. So it may be that their switch list just hasn't been updated.

Plus we did once have this in the json but then later dropped it as per the changelog:
https://github.com/rockstor/rockon-registry/commit/dca3235dd672e63af2b0f45157fe44b61a3e0b9e#diff-f702dd7f333fdcf0a19164394e0bedf0

I only remembered as it was the first time we had a volume removed from a json and I had to look to a bug that surfaced as a result.

alazare619:
While not 100% required. It most definitely does infact work. If for instance you have a seperate share on an SSD you can map the /transcode and then in the plex console tell it the directory is /transcode. It nots 100% required and I'm not certain how to add a share via json thats not required.

phillxnet:
"I'm not certain how to add a share via json thats not required." This 'feature' is already enabled on the Plex rock-on, ie see the 'i' icon:
"You can add more Shares(with media) to Plex from the settings wizard of this Rock-on. Then, from Plex WebUI, you can update and re-index your library."
It's under the spanner icon "Add Storage" button:

With pointers and pic in our:
"Adding Shares to Plex" doc subsection of "Plex Server Rock-on" of "Rock-ons with write-ups":
http://rockstor.com/docs/docker-based-rock-ons/plex-media-server.html#adding-shares-to-plex
and the json element that enables this for any rock-on is covered in the README of this repo:
```
(optional)"volume_add_support": true, If the app allows arbitrary Shares to be mapped to the main container>,
```
Just tried it and once added appears as such:

plex-transcode-dir

As such my personal preference would be to not complicate the Plex setup with 3 share choices and stick to our current 2. Though I think opening a doc issue to enhance the quoted section to indicate it's potential use for this rather advanced setup would make for a great addition. We could also include there how this directory can be selected, ie PlexUI -Spanner (settings)- Server - Transcode - Show Advanced. Along with 'ssd recommended' and the like and maybe some more advanced mount options or the like.

My experience just now (quick test) was (as you say) that it did not appear to 'by default' make use of the /transcode directory until I configured it to do so via the above PlexUI setting. And once configured (within PlexUI) the share has a "Transcode" directory created within it by Plex the first time it's called upon so that's nice.

All good stuff and hopefully that means what we already have this capability without complicating things for the less advanced users.

So good call but I favour doc additions to keep simple the initial setup, especially given we can already accommodate this kind of custom mapping.

Hope that helps.

alazare619:
Sounds like a sound idea I'll close this. Is there a place to do a doc for this or would you rather I just do a pr w/ markdown format?

phillxnet:
@alazare619 OK. As per doc it's entirely up to you, you can either do a pr on that plex doc which is hosted in our rockstor-doc repo:
https://github.com/phillxnet/rockstor-doc
and specifically:
https://github.com/phillxnet/rockstor-doc/blob/master/docker-based-rock-ons/plex-media-server.rst
it's a markdown of sorts but actually restructured text.
And our docs themselves have a section on how to contribute, including testing the format via Sphinx:
http://rockstor.com/docs/contribute_documentation.html

The docs as presented at http://rockstor.com/docs/ are essentially Sphinx rendered html from source rst and png files in that repo, though not directly.

Or you could do a forum thread or wiki entry:
https://forum.rockstor.com/

I would favour the doc enhancement but then it's not me doing it. Either way it's a useful addition and given your familiarity with Git etc you might find the docs pr entry just as easy. We currently only have one issue open on the plex doc and it's on an unrelated element.

See how you feel and take it from there. I think it would make a great addition where ever and am considering doing this on my own installs so look forward to the guide with what ever tips you can throw in there: ie size recommendations etc. Also don't shy away from linking to upstream docs as ultimately they are the 'source of truth'.

Thanks for stepping up for this one and I look forward to the read.

phillxnet:
@alazare619 as per your "Sounds like a sound idea I'll close this." from a time now long forgotten I will close this pr in accordance with our prior discussion. Thanks for the offer to improve the docs by the way and as usual any help you can provide is most welcome.

Closing as per prior agreement (and to tidy up our pr backlog), please re-open if I've misread something.

PR #142, opened 2018-01-16
Unms

alazare619:
Ubiquiti Network Monitor System. This is the future replacement to the Ubiquiti Unifi suite. This supports Airmax, Edgemax (Edgerouters, Switches etc) and the Unifi gear.

### DIFF
--- a/unms.json
+++ b/unms.json
@@ -1,0 +1,46 @@
+{
+    "UNMS": {
+        "containers": {
+            "unms-oznu": {
+                "image": "docker pull oznu/unms",
+                "launch_order": 1,
+                "ports": {
+                    "80": {
+                        "description": " Expose the HTTP web server port on the docker host, Leave Default of 80",
+                        "host_default": 80,
+                        "label": "http80",
+                        "ui": false
+                    },
+                    "443": {
+                        "description": "Expose the HTTPS and WSS web server port on the docker host, Leave default 443",
+                        "host_default": 443,
+                        "label": "https443",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a Share for UNMS configuration. Eg: create a Share called UNMS-config for this purpose alone.",
+                        "label": "Config Storage"
+                    }
+                },
+                "environment": {
+                    "PUID": {
+                        "description": "Enter a valid UID to run UNMS with. It must have full permissions to all Shares mapped in the previous step.",
+                        "label": "UID",
+                        "index": 1
+                    },
+                    "PGID": {
+                        "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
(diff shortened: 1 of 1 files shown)


phillxnet:
@alazare619 Not up enough on this software to comment on it's rock-on roll really but I would say to take care with adding a '#' char followed by numbers in commit messages, as you have done in the commit starting with:

"flip flop ports so ui button works attempt"

as GitHub defaults to assuming you want to link to that issue or pr in the same repo. Thus you have linked to the very first issue in the rockon-registry which I'm assuming was not your intention. Your only the 3rd to do this so look like it will be an oft referenced issue :).

Thanks for you efforts on this by the way. Is it ready for review yet? I.e. have you tested it as is?
Will be another great addition once in.

alazare619:
I'm running it at the moment its managing my edge router x so far
flawlessly running it in conjunction with the unified controller the only
gotcha is unms wants 443 and 80 for their ports

phillxnet:
@alazare619 Thanks again for yet another Rock-on and again sorry for the unreasonable delay.
As per my prior messages I'd like to get this in if it's still workable.

My only concern now is the relevance of your:
"... the only gotcha is unms wants 443 and 80 for their ports."
We should really have something in the "description" section that warns users that they are going to have  reconfigure Rockstor's Web-UI in order to run this Rock-on.

Requesting a review of both my change request and this pending pr.

Hooverdan96:
@alazare619, @phillxnet, @FroggyFlox 
For this PR unfortunately the image utilized has been archived by the maintainer. I found this _all-in-one_ container that seems fairly recent and also offers multiple architectures, amd64 and arm (**nico640/docker-unms
**). Could this one be shoehorned in here? It has a few optional environment variables, I suspect two of the optional ones are relevant if one experiments with other ports than the default ones:
-e PUBLIC_HTTPS_PORT=443 - This should match the HTTPS port your are exposing to on the docker host
-e PUBLIC_WS_PORT=443 - This should match the HTTPS port your are exposing to on the docker host
It would still mean that the "same" port has to be maintained twice (once in the minimum required parameters and once with the optional ones here), but it seems a necessary evil at this time.
What do you think?

phillxnet:
@alazare619 & @Hooverdan96, I'd like to see the work here end up in a Rock-on.
any change we can pick this one up. As @Hooverdan96 points out It seems the image used is archived. Can we get this pr updated with a new multi-arch image possibly. Or are we looking at a core change requirement.

I'd like to clear out our backlog of potentially rusty pull request and I suspect this one has slipped through our fingers unfortunately.

Hooverdan96:
when I get a chance I take another look at it, unless @alazare619 has time to resolve this one.

phillxnet:
@FroggyFlox and @Hooverdan96 We have had no takers on the review of this pr.
Should be now close? In light of our new close old (with not activity) stuff and delete existing stuff if broken via my proposed new policy.

PR #144, opened 2018-02-03
Logitechmediaserver

jfearon:
Remove the old version of logitech squeezebox server and replace with docker image for logitech media server.  This means that:

1) Logitechs rename is reflected in the rock on name
2) by switching docker image the latest version of the media server is pulled in


jfearon:
Not sure about the filename change - if that causes massive problems it can be changed!

Brookke:
@jfearon looks like you need to also update the root.json file 😊

phillxnet:
@jfearon Thanks for your earlier efforts on this. I'm doing some tidy up and this looks to be in limbo. Would be nice to have it reach a conclusion of sorts.
I appreciate that much time has passed but if this is still of interest then @FroggyFlox last comment might be a good place to start.

phillxnet:
Closing as resolved in #300 and #283 respectively.

PR #148, opened 2018-06-06
Update Unifi rockon

sgissi:
- Move UI port from 8080 to 8443
- Removed port 8081, no longer being used
- Added Speed Test and UBNT discovery ports
- Volumes are now under /unifi instead /var/lib/unifi and contains configuration and logs (upstream change)
- Run as specific user instead of root

### DIFF
--- a/unifi.json
+++ b/unifi.json
@@ -10,21 +10,14 @@
                         "host_default": 8080,
                         "label": "Redirect to WebUI",
                         "protocol": "tcp",
-                        "ui": true
+                        "ui": false
                     },
                     "8443": {
                         "description": "Port for WebUI and API. Suggested default: 8443",
                         "host_default": 8443,
                         "label": "HTTPS WebUI",
                         "protocol": "tcp",
-                        "ui": false
-                    },
-                    "8081": {
-                        "description": "Management Port. Suggested default: 8081",
-                        "host_default": 8081,
-                        "label": "Management Port",
-                        "protocol": "tcp",
-                        "ui": false
+                        "ui": true
                     },
                     "8880": {
                         "description": "Redirect to Portal. Suggested default: 8880",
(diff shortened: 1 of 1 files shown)


sgissi:
Ping @magicalyak for review :) Upstream modifications of the Docker image, this PR will bring the Rockon up to speed and more secure now that we can run without root.

magicalyak:
One issue I did find, if you change a port it messed up since the port mappings are hardcoded.  Any thoughts? 
6789 for me is 6791 (the port is used by NZBGet)

magicalyak:
Nevermind I think I just need to change the conf file in the config directory, the container looks fine.

magicalyak:
@sgissi I see linuxserver has a really popular unifi container.  I have a working json for it.  I also see that if we update this json it uses a different directory structure so there isn't a clean upgrade for it (maybe just add a note?)  Honestly the end user would get a new install and they can just import a backup.  
Here is my working one - https://gist.github.com/magicalyak/6625ce8c22dad5bcbe9ecbef48ab514e

sgissi:
@magicalyak Both seems to be quite up-to-date with the releases and I understand linuxserver already sustaining other containers. The only drawback I can see is that 5.7.x train (which I use currently) is labeled as "unstable" on linuxserver and the specific version is not clear to me. If we use linuxserver default image then existing users might lose functionality and/or have issues with backup/restore between 5.6 and 5.7. Maybe break into two Rockons? Unifi = linuxserver/unifi:unstable to avoid breaking existing users and Unifi Stable using linuxserver/unifi? I'm sure risk averse users would gravitate towards "Unifi Stable" given the name :)

sgissi:
@magicalyak Any feedback on this one? I can open a PR to create two Unifi entries like I proposed before if you are Ok with it (or "Unifi" and "Unifi Unstable" but I don't like the connotation, 5.7.x has been pretty good).

magicalyak:
@sgissi I think make this the 3.7 and I can use unifi-linuxserver as the stable one.  The updates here actually break existing unifi installs because the config path changes to a nested directory.

sgissi:
Fair point. I'll spin up a 3.7 with the existing Unifi image and layout, then run the new Rockon I'm proposing here with the same data share, just to make sure the upgrade is safe.

phillxnet:
@sgissi Thanks for your continued maintenance on this Rock-on, and apologies for the some what massive delay. But I'd like to know what the status is now of this pr?

From your last comment it looks like you have in mind a final test regarding migration.

Let me know what you want to do and upon a confirmation of your changes (ie a review from anyone familiar with this software) I'd be chuffed to do the merge. And not breaking existing rock-on installs would be great so anything you can do in that direction is likely to be much appreciated :)

@magicalyak as always thanks for you valuable and valued input on this one.

phillxnet:
@sgissi @magicalyak I'm wondering if this pr is now redundant given our merging of:
"updated unifi-linuxserver" #192

Should be maybe drop this json entirely?

I not at all familiar with this software and we have multiple versions of other offerings.

Thoughts and more informed opinion welcome as always. There was some discussion relating to migrating folks from pre this pr to post? And that would seem relevant if we are to keep this offering. I.e. current users of this Rock-on.

magicalyak:
Yeah the other one should be primary at this point.

phillxnet:
> Yeah the other one should be primary at this point.

Thanks. So what do you think of just removing the json associated with this pr. We are well overdue for a tidy up and having a load of unmaintained Rock-ons, especially when they are redundant, may well be the way to go. And in this case a backup restore of settings would seem like a possibility if folks are using this one and need to re-install.

@FroggyFlox calling also on your opinion here re pruning. Maybe we should have a stint of that for a change. Fewer better supported Rock-ons is better in my opinion; especially in the light of redundancy: ie duplicate function/project newer Rock-ons.

magicalyak:
Remove it is my vote

phillxnet:
@magicalyak 
> Remove it is my vote
Agreed. @FroggyFlox do you see a problem with us removing the json and root entry associated with this pr. Given it's fallen out of repair and we now have a replacement I think it would be a better all round user experience if we at least do pruning from time to time.

FroggyFlox:
> @FroggyFlox calling also on your opinion here re pruning. Maybe we should have a stint of that for a change. Fewer better supported Rock-ons is better in my opinion; especially in the light of redundancy: ie duplicate function/project newer Rock-ons.

@phillxnet, I would tend to agree with @magicalyak and vote to remove this one if both offer the same version of Unifi and features. Note that I'm unfamiliar with the software itself so I'll rely on the discussion above on that.
Note that while both images (jacobalberty and lsio) seem to be kept up-to-date rather well (at least as of today), the lsio ones do offer special features for customizations that could be useful down the road. LSIO is also smaller:
- jacobalberty: 307 MB
- lsio: 190 MB

The question would be how to deprecate this image, however:
- should we have a transition period with some warning in the webUI?
- remove entry from `root.json` but keep the rock-on json file on the repo for users who want to keep using it via local metastore?

I've never tested it, but based on my understanding, if the entry were to be removed from `root.json`, clicking "Update" in Rockstor UI will not make the rock-on disappear as long as it is in "installed" state. If the user uninstalls it, however, it will be remove from the list of available rock-ons.

Hope this helps,

sgissi:
Ops, just an year later :) I have moved my Unifi setup out to Kubernetes and didn't came back to close the PR. I agree it can be removed.

FroggyFlox:
No worries, @sgissi , thanks a lot for taking care of this!

PR #150, opened 2018-07-01
FileRun Rockon

magicalyak:
This rockon is for FileRun - http://filerun.com/  - Self-Hosted File Sync and Sharing
The container is based off a docker repo that uses linuxserver-like setup (it uses their xenial base and layer but it's maintained by them).

### DIFF
--- a/filerun.json
+++ b/filerun.json
@@ -10,7 +10,6 @@
                         "description": "FileRun WebUI port. Suggested default: 8080",
                         "host_default": 8080,
                         "label": "WebUI port",
-                        "protocol": "tcp",
                         "ui": true
                     }
                 },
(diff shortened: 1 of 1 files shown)


magicalyak:
Don't merge until I have confirmation on the full setup.  The rockon currently seems to require a mysql that isn't included.  I will add to this a 2-container pull or instructions on using another rockon to provide the database.

phillxnet:
@magicalyak I'll close this for now, by way of tidy-up as it seems to have slid into the realms of having received not attention.

My apologies the to stop-start nature we have had in this repo, but our new approach of deleting/closing old & un-attended requests and existing Rock-ons should help with seeing he wood for the trees.

Again thanks for all you effort over the years here. Much appreciated.

PR #151, opened 2018-07-27
Update ghost.json

guppy0130:
Add fix for breaking change as described by https://github.com/docker-library/ghost/issues/69


phillxnet:
@guppy0130 Thanks for the fix, much appreciated and nice to have the upstream link included.

Can you confirm that the ghost Rockon now works as expected?

Cheers.

guppy0130:
Installed on my local rockstor installation and appears to be working as expected.

Issue #31, opened 2016-04-03
Make Rock-on links consistent


phillxnet:
When clicking on the names of Rock-ons I think it should link to the upstream project, not the docker maintainers page. This way users will get to see what they are to install not the details of the packager / docker container.
However it would also be nice to acknowledge the origin of the docker container so I propose to normalize this info by having the subtitle of each Rock-on name via a "by" entry the docker maintainer. This is already done in for example the current Plex Rock-on but in that same case we don't link to the Plex site via the name.
Summary:
Names link to upstream projects ie as with Syncthing currently.
subtitles (below names) identify that the system is "Continous File Synchronization" but optionally adds "docker image by ......". This way we cover both options where the more technical can search for the docker maintainer but the general title link serves the initial upstream link to the main project.

phillxnet:
Can we have input on this idea as I still think it stands up. Plus it would good to get this expected behaviour in place sooner rather than later. Especially as our current list of Rock-ons is now quite large.

phillxnet:
Including a quote from 'closed as duplicate' issue by @seffyroff towards this discussion:

_"Right now browsing the Catalog of Rockons is very sparse. I don't know the source of the Rockon (whose Plex comtainer is that) without looking in Github at the contents of the .json. I don't know the version number, or last update version - these are important :)"_

Thanks @seffyroff .

FroggyFlox:
Hi @phillxnet ,

I think that's a good idea and does make more sense as it allows the user to get more information on what this Rock-on actually does. I also agree that we need to keep a good description of the image's particularities especially as we now have more and more projects with multiple Rock-ons (Owncloud for instance).
Should we push for a more homogenous use of the `"description:` object in the JSON file, with some sort of quick template?
We also have the `"version"` object that could be used further to help alleviate this problem.

phillxnet:
I think all descriptions should contain example share names so it just makes for an easier experience. So definitely that. I think all the early ones did have this and it got missed as we went on.
Also keen on formalising within the json as then all new rock-ons can be vetted to contain this and we can problematically extract with more reliability. Potentially properly offering a Web-UI component that reflects this element as an auto create if non existent option with a warning if it already exists. This would also make redundant the bespoke inclusion of this info within a description.

As to the version element. Yes that one has seen some attention; see:
[Enhancement - Rockons] Add current Rock-on version on Rock-ons table
https://github.com/rockstor/rockstor-core/issues/1636
where @MFlyer also submitted a pr:
[...]

PR #155, opened 2018-08-14
Add nginx rockon

JasonTarka:
Use the official nginx image, but users will need to add a `default.conf` file to their nginx config share for the server to work. This is required as the mounted volume will shadow the `config` directory in the container that nginx tries to read the config file from.

### DIFF
--- a/nginx.json
+++ b/nginx.json
@@ -1,0 +1,44 @@
+{
+	"nginx": {
+		"description": "Open source reverse proxy server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a web server.",
+		"version": "1.0.0",
+		"website": "https://www.nginx.com/",
+		"icon": "https://cdn-1.wp.nginx.com/wp-content/themes/nginx-theme/assets/img/logo.svg",
+		"more_info": "At startup the config file named default.conf gets created in the config volume if the directory is empty. Extend it with your own custom configuration.<br/>An example configuration file can be found <a href='https://www.nginx.com/resources/wiki/start/topics/examples/full/'>in the nginx wiki</a>. The config volume can be refered as <pre>/config</pre>",
+		"containers": {
+			"nginx": {
+				"image": "jtarka/nginx",
+				"tag": "latest",
+				"launch_order": 1,
+				"ports": {
+					"80": {
+						"description": "HTTP port",
+						"host_default": 1080,
+						"label": "HTTP port",
+						"protocol": "tcp",
+						"ui": true
+					},
+					"443": {
+						"description": "HTTPS port.",
+						"host_default": 1443,
+						"label": "HTTPS port",
+						"protocol": "tcp",
+						"ui": true
+					}
+				},
+				"volumes": {
+					"/config": {
+						"description": "Choose a Share for the nginx configuration. Eg: create a Share called nginx-config for this purpose alone.",
+						"label": "Config Storage",
+						"min_size": 10240
+					},
+					"/usr/share/nginx/html": {
+						"description": "Choose a Share for the files that nginx will serve (HTML, JavaScript, etc.). Eg: create a share called nginx-data for this purpose alone.",
+						"label": "Served Content Storage",
+						"min_size": 102400
+					}
+				}
+			}
+		}
+	}
+}
--- a/nginx.json
+++ b/nginx.json
@@ -1,10 +1,10 @@
 {
 	"nginx": {
-		"description": "Open source web server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a reverse proxy server.<p>Advanced users only: You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388'>add a default config file to your "Config Storage" share.</a>.</p><p>Based on the official Nginx docker image: <a href='https://hub.docker.com/_/nginx' target='_blank'>https://hub.docker.com/_/nginx</a>.</p>",
+		"description": "Open source web server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a reverse proxy server.<p>Advanced users only: You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388' target='_blank'>add a default config file to your Config Storage share.</a>.</p><p>Based on the official Nginx docker image: <a href='https://hub.docker.com/_/nginx' target='_blank'>https://hub.docker.com/_/nginx</a>.</p>",
 		"version": "1.0.0",
 		"website": "https://www.nginx.com/",
 		"icon": "https://cdn-1.wp.nginx.com/wp-content/themes/nginx-theme/assets/img/logo.svg",
-		"more_info": "You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388'>create a default config file</a> in the Config Storage share before starting this Rockon, which you can extend with your own custom configuration.<br/>For advanced users, an example configuration file can be found <a href='https://www.nginx.com/resources/wiki/start/topics/examples/full/'>in the nginx wiki</a>. The config volume is mounted at <pre>/etc/nginx/conf.d/</pre>",
+		"more_info": "You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388' target='_blank'>create a default config file</a> in the Config Storage share before starting this Rockon, which you can extend with your own custom configuration.<br/>For advanced users, an example configuration file can be found <a href='https://www.nginx.com/resources/wiki/start/topics/examples/full/' target='_blank'>in the nginx wiki</a>.",
 		"containers": {
 			"nginx": {
 				"image": "nginx",
@@ -23,7 +23,6 @@
 						"host_default": 1443,
 						"label": "HTTPS port",
 						"protocol": "tcp",
-						"ui": true
 					}
 				},
 				"volumes": {
--- a/nginx.json
+++ b/nginx.json
@@ -1,6 +1,6 @@
 {
 	"nginx": {
-		"description": "Open source web server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a reverse proxy server.<p>Advanced users only: You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388' target='_blank'>add a default config file to your Config Storage share.</a>.</p><p>Based on the official Nginx docker image: <a href='https://hub.docker.com/_/nginx' target='_blank'>https://hub.docker.com/_/nginx</a>.</p>",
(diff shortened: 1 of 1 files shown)


magicalyak:
Jason, I don't know if this was an old one or one I did before but I had this in my testing folder.  It's linuxserver based (which we prefer due to stability) and if you can verify it, feel free to take credit and submit it.

{
    "nginx": {
        "containers": {
            "nginx": {
                "image": "linuxserver/nginx",
                "launch_order": 1,
                "ports": {
                    "80": {
                        "description": "HTTP port. Suggested default: 80",
                        "host_default": 80,
                        "label": "HTTP port"
                    },
                    "443": {
                        "description": "HTTPS port. Suggested default: 443",
                        "host_default": 443,
                        "label": "HTTPS port"
                    }
                },
                "volumes": {
                    "/config": {
                        "description": "Choose a Share for configuration. Eg: create a Share called nginx-config for this purpose alone.",
                        "label": "Config Storage"
                    }
                },
                "environment": {
                    "PUID": {
                        "description": "Enter a valid UID to run with. It must have full permissions to all Shares mapped in the previous step.",
                        "label": "UID",
                        "index": 1
                    },
                    "PGID": {
                        "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
                        "label": "GID",
                        "index": 2
		                }
                }
            }
        },
        "description": "nginx by Linuxserver.io",
        "more_info": "This Container is a simple nginx webserver configured with default and ssl, and all relevant config files moved out the user via the config directory for ultimate control. it contains some of the basic php-packages. and is built on our internal nginx baseimage.",
	      "volume_add_support": true,
        "website": "https://hub.docker.com/r/linuxserver/nginx/",
        "version": "1.0"
    }
}

nfriedly:
I just coppied @magicalyak's config and it seems to be working for me. I mapped a different share to /config/www to separate my config from my content.

I'll also mention that there is now an official docker nginx image: https://hub.docker.com/_/nginx/ 

The Linuxserver.io image includes php support, so it might make more sense to name it something like php-nginx and have both available as separate rock-ons.

phillxnet:
@JasonTarka Thanks for the Rock-on submission and apologies for the significant delay.
I'm happy to merge this but I think @nfriedly has a good point in that if there is now an official nginx docker image then it should have the plain nginx name. Let me know if I've missed something obvious here however.

@magicalyak Thanks for yet another review, fancy submitting your json via another pr and giving it an appropriate linuxserver name with @nfriedly php suggestion added maybe?

I'm going to label this "help needed" as we need a consensus on where to go with it.

My suggestion is we name this one after it's image at least ie:
"nginx-jtarka"
however if we have other Rock-ons with more popular images then maybe we should keep duplication to a minimum.

Suggestions welcome.

magicalyak:
@phillxnet you want me to submit a new PR based on linuxserver.io or should we use this one?  JsaonTarka, we try to avoid using self-managed repos when possible but it's not a hard rule.  Also, we've organically adopted linuxserver.io repos when possible (this isn't official but is a trend).  For copying files, I use Dockerfile and something like an entrypoint.sh  if you want take a look at magicalyak/pocketmine (I'm honestly having some issues with the program but the dockerfile and entrypoint.sh work fine.  You can see it copies server.properties and even downloads and unzips a file).  The other container that does this is the letsencrypt one I think.  I'd use these as examples.  The json won't copy files but the dockerfile and scripts it calls certainly can.

phillxnet:
@JasonTarka Thanks for the clarification and chuffed you are still up for contributing. And yes I think we do care about providing a default config as I think many users would appreciate an 'up and running' variant to a blank / broken one. And if they can deal with a blank variant then they can also deal with a skeleton one, but the reverse is not true, ie greater target audience if we are up and running directly after install. Oh and re image selection: what @magicalyak said as they have greater experience than me in building docker images and authoring Rock-ons. However given @magicalyak offer above I think we can have both options if you are game. See later in this comment.

@magicalyak  Thanks for you input on this one. Much appreciated.
Yes I think submitting an appropriately named linuxserver.io image based nginx Rock-on would be great as we know where we are with those and they have a well deserved reputation. So if you want to pop that in with an obvious name then that would be great. And for those that know less of what image (Rock-on in this case) to chose I again second @nfriedly's suggest that we emphasise either in name or more appropriately in description that it's PHP capable as that will be a key indicator of choice for some.

@JasonTarka So given @magicalyak looks to be up for rolling our 'friendly' variant of nginx based on a well know docker image, maybe there is also a place for an 'advanced' official variant but with a clear indication in the Rock-on description that it is 'broken' out of the door. I.e:

"This official nginx docker image based Rock-on has no default configuration and will therefore not work until one is provided within the given share (show-shares-root-mount-point-here ). Advanced nginx admin use only."

type thing; or what ever better wording you can come up with.

That way, if you are game, we can span the gamut. @magicalyak's friendly batteries included linuxserver.io variant and your vanilla 'no config' variant for those that wish to start from scratch using the official docker image only.

Are we all agreed that these two options can co-exist as long as we name and describe them for what they are?

magicalyak:
Good with me, I'll create a new nginx for the linuxserver.io

JasonTarka:
If you're okay with it not working out of the box, then I'm okay with it using the official nginx image. I can completely understand not wanting to use some random person's containers, if for no other reason than the potential security issues it could cause.

I'll update this PR to change the image name, and make the description tell the user what they need to do, including a link to a sample config file they can drop in.

phillxnet:
@JasonTarka Thanks for your efforts on this and for accommodating the 'advanced' end of our nginx user base. Would be great to put this one to bed.

nginx.json
```diff
@@ -0,0 +1,44 @@
+{
+	"nginx": {
+		"description": "Open source reverse proxy server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a web server.",
```
phillxnet:
```suggestion
		"description": "Open source web server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a reverse proxy server.<p>Advanced users only: You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388'>add a default config file to your "Config Storage" share.</a>.</p><p>Based on the official Nginx docker image: <a href='https://hub.docker.com/_/nginx' target='_blank'>https://hub.docker.com/_/nginx</a>.</p>",
```

phillxnet: @JasonTarka I know it's been some time but we now have a different format we are attempting to add to all existing and future Rock-ons. Linking to the docker image used is part of that format.

nginx.json
```diff
@@ -0,0 +1,44 @@
...
+						"description": "HTTPS port.",
+						"host_default": 1443,
+						"label": "HTTPS port",
+						"protocol": "tcp",
+						"ui": true
+					}
+				},
+				"volumes": {
+					"/etc/nginx/conf.d": {
+						"description": "Choose a Share for the nginx configuration. Eg: create a Share called nginx-config for this purpose alone.<br/>You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388'>add a default config file</a>.",
```
phillxnet:
```suggestion
						"description": "Choose a Share for the nginx configuration. Eg: create a Share called nginx-config for this purpose alone.",
```

phillxnet: I've moved the link for your example config from the share description to the general description as links are not processed in the tooltip used against each share.

phillxnet:
@magicalyak It's been a while but I find myself in this pr again and thought I'd drag up the following prior chat of our:

> @magicalyak Thanks for you input on this one. Much appreciated.
> Yes I think submitting an appropriately named linuxserver.io image based nginx Rock-on would be great as we know where we are with those and they have a well deserved reputation. So if you want to pop that in with an obvious name then that would be great. And for those that know less of what image (Rock-on in this case) to chose I again second @nfriedly's suggest that we emphasise either in name or more appropriately in description that it's PHP capable as that will be a key indicator of choice for some.

Might be nice to extend our 'developer' Rock-ons range.

phillxnet:
@JasonTarka we have had a potential renewed interest in this pull request from @mikey0000 in the following issue:
https://github.com/rockstor/rockstor-core/issues/2129
Let us know if you are game to pick this up again with my recent suggestions otherwise we can start over in a new pr from where I unfortunately neglected this pull request from way back.

JasonTarka:
This is outside the scope of this PR, but is the standard now to manually include a link to the image in the description?

It feels like that's something that could (and probably should) be pulled dynamically from the `containers` config. Generating the links would improve consistency (especially visual), reduce manual effort, and lessen the chances of the links becoming stale if the selected image changes.

FroggyFlox:
> This is outside the scope of this PR, but is the standard now to manually include a link to the image in the description?


You are correct; see Issue #31 for the history and origin for this.


> It feels like that's something that could (and probably should) be pulled dynamically from the `containers` config. Generating the links would improve consistency (especially visual), reduce manual effort, and lessen the chances of the links becoming stale if the selected image changes.


That is an excellent idea and something that should be implemented. Would you mind opening an issue for that in the [rockstor-core repository](https://github.com/rockstor/rockstor-core/issues) (as this is where the changes would need to be made)? This way we can keep track of it and make it available for somebody to start working on it if desired.

nginx.json
```diff
@@ -0,0 +1,44 @@
...
+						"label": "HTTP port",
+						"protocol": "tcp",
+						"ui": true
+					},
+					"443": {
+						"description": "HTTPS port.",
+						"host_default": 1443,
+						"label": "HTTPS port",
+						"protocol": "tcp",
+						"ui": true
```
phillxnet:
```suggestion
```

nginx.json
```diff
@@ -0,0 +1,44 @@
+{
+	"nginx": {
+		"description": "Open source web server for HTTP, HTTPS, SMTP, POP3, and IMAP protocols, as well as a load balancer, HTTP cache, and a reverse proxy server.<p>Advanced users only: You must <a href='https://gist.github.com/JasonTarka/63d632b5633fb0840cdfb3b56a14d388'>add a default config file to your "Config Storage" share.</a>.</p><p>Based on the official Nginx docker image: <a href='https://hub.docker.com/_/nginx' target='_blank'>https://hub.docker.com/_/nginx</a>.</p>",
```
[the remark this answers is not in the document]
JasonTarka: Easy to miss, though I'd have thought there'd be some tests to catch that. If I find the time, I'll see what needs to be done to add some.

JasonTarka:
> > It feels like that's something that could (and probably should) be pulled dynamically from the `containers` config. Generating the links would improve consistency (especially visual), reduce manual effort, and lessen the chances of the links becoming stale if the selected image changes.
> 
> That is an excellent idea and something that should be implemented. Would you mind opening an issue for that in the [rockstor-core repository](https://github.com/rockstor/rockstor-core/issues) (as this is where the changes would need to be made)? This way we can keep track of it and make it available for somebody to start working on it if desired.

Created an issue to track this: rockstor/rockstor-core/issues/2131

magicalyak:
I just realized you had pinged my on this @phillxnet and it's been a while, is the nginx proxy rockon sufficient for our needs at this point or should i push forward with the linuxserver.io one?

phillxnet:
@magicalyak Hello again. Re:

> ... is the nginx proxy rockon sufficient for our needs at this point or should i push forward with the linuxserver.io one?

I think it would be great to have both this one, nearly done, and a beginner friendly linuxserver.io one as the latter would presumably run as selectable non root user with default config and PHP and what not. This one servers better those that want to use the official nginx docker image and their own or @JasonTarka example config.

So if you are still game then it would be great to also have a linuxserver.io nginx one, we just need to name it accordingly as discussed earlier in this thread.

Thanks for the response and being willing to submit yet another Rock-on.

phillxnet:
@JasonTarka, @FroggyFlox, @magicalyak, and @nfriedly 

I've now added commit bb584d291cb56fd0d45417e13993fafadf8c3053 to complement this pr's now merged json and am about to publish.

Thanks to all.

PR #156, opened 2018-08-15
Update to correct repo (and latest version)

laxdog:
The repo currently being used for this rockon is deprecated and only hosts the 3.X versions. 

This is the new updated repo and has the 4.X versions.

### DIFF
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -1,8 +1,8 @@
 {
     "Pi-Hole": {
         "containers": {
-            "pi-hole-diginc": {
-                "image": "diginc/pi-hole",
+            "pi-hole": {
+                "image": "pihole/pihole",
                 "opts": [
                     [
                         "--cap-add",
@@ -71,13 +71,13 @@
                 }
             }
         },
-        "description": "PI-Hole by DigInc",
+        "description": "The official Pi-hole Docker image",
         "more_info": "<h4>PI-HOLE™: A BLACK HOLE FOR INTERNET ADVERTISEMENTS</h4><p><b>Admin page</b></p><p>To access admin interface go to URL: http://[SERVERIP]/Admin</p><p>If you have different port than 80 you need to specify that in the URL.</p><p><b>Block Over 100,000 Ad-serving Domains</b></p><p>Known ad-serving domains are pulled from third party sources and compiled into one list.</p><p><b>Block Advertisements On Any Device</b></p><p>Network-level blocking allows any device to block ads, regardless of hardware or OS.</p><p><b>Improve Overall Network Performance</b></p><p>Since ads are blocked before they are downloaded, your network will perform better.</p>",
         "ui": {
             "slug": "admin"
         },
         "volume_add_support": false,
-        "website": "https://hub.docker.com/r/diginc/pi-hole/",
-        "version": "1.0"
+        "website": "https://hub.docker.com/r/pihole/pihole/",
+        "version": "1.1"
     }
 }


laxdog:
I've added the TZ, pulling the actual timezone from the OS. I'm having trouble testing it now though as it won't install and I can't find any logs. I'll keep poking it.

As for the dnsmasq, it does seem useful and in fact I missed that in the docs when I was doing the update. 

How will that work though if there's a new user input required? If people have it already installed and update their rockons what happens?

FroggyFlox:
Sorry to read it won't install now. I think your idea of pulling the time zone from the OS is a better idea. One less variable to set at install is always an improvement.

Regarding the update question, the user will have to uninstall before "updating" their list of Rock-on to fetch the new JSON and then reinstall the Pi-hole Rock-on. They will then be able to reuse their old pihole-conf share and see they now need another share for dnsmasq. At least that's how I believe it should take place.

magicalyak:
> Another little thing I forgot to discuss: the restart policy.
> 
> It's my understanding that with the current `--restart=always` setting, the Rock-on will be automatically turned back on upon a reboot of the computer. If I'm correct, it may be preferable to set it to `--restart=unless-stopped` as in the container's doc to respect the user's choice of turning the Rock-on off.
> 
> Any thought?

I agree, Rockstor handles the restart options automatically so this should be removed.

haioken:
FYI - Ran up a pi-hole instance with this Rockon from scratch and needed to specify "docker run" args:
--dns 127.0.0.1
--dns 8.8.8.8

Without these flags, pi-hole was unable to perform any DNS lookups, and the UI would not start.
Reason from docker logs was:
dns resolution is currently unavailable

phillxnet:
@laxdog Thanks for the submission and I know it's been a while but do you fancy picking this pr up again.
I am now able to merge in the repo and would like to help but it looks like this proposed Rock-on need just a little more attention re the comment from @FroggyFlox and @magicalyak, thanks to the both of your for you input.
@haioken Thanks for your review also much appreciated.

So is the consensus that we need the time zone fix, the explicit restart policy removed, and additional dns arguments as per @haioken's comments?

FroggyFlox:
> So is the consensus that we need the time zone fix, the explicit restart policy removed, and additional dns arguments as per @haioken's comments?

Yes. I would also add the `/etc/dnsmasq.d/` volume as mentioned above, as it is an important addition to me (making dnsmasq configuration files persistent). It is also a setting recommended on the image's documentation.
Most importantly, @haioken suggestions **need** to be implemented as otherwise it throws a blocking error.

magicalyak:
I can't find the original branch for him, I could make a new one with the change @haioken said.  Standby...Probably a new PR.
@laxdog please let me know if you want to fix this on this PR, I always feel like i'm stealing someone's thunder opening a new one.

haioken:
@magicalyak - It looks like this @FroggyFlox's request for the dnsmasq volume isn't in this PR - any particular reason for it?
This is something I'd like to see as well, I have my dodgy fix from the forums, but having it persist across multiple restarts or updates would be preferable.

magicalyak:
I’ll add it. Just didn’t yet. Might be tomorrow though. I want to review the options again and make sure we have everything.

magicalyak:
@haioken - check out and comment on https://github.com/rockstor/rockon-registry/pull/178
This should work.

phillxnet:
As per @magicalyak last comment, this pr is now obsoleted (Duplicate) by the now merged pr of #178.
Thanks to @magicalyak , @FroggyFlox , @haioken and @laxdog.

We got there in the end.

PR #157, opened 2018-08-26
Add netdata Rockon

FroggyFlox:
Although Rockstor's dashboard provides a quick and instantaneous overview of most of the server's state, it could be limited for some users who would like to have a more detailed view of the machine's CPU, RAM, DISK, and NET usage.
Its documentation reads:

> netdata is a scalable, distributed, real-time, performance and health monitoring solution for Linux, FreeBSD and MacOS. It is open-source too.
> 
> Out of the box, it collects 1k to 5k metrics per server per second. It is the corresponding of running top, vmstat, iostat, iotop, sar, systemd-cgtop and a dozen more console tools in parallel. netdata is very efficient in this: the daemon needs just 1% to 3% cpu of a single core, even when it runs on IoT.

Notably, netdata allows the users to set specific alerts at custom thresholds, which could be of great interest for some. This would require, however, the user to mount an additional share and add email credentials that most users may not need/want to add. As a result, these may be better suited for use out of Rockons.

### DIFF
--- a/netdata.json
+++ b/netdata.json
@@ -1,0 +1,40 @@
+{
+    "Netdata": {
+        "containers": {
+            "netdata": {
+                "image": "titpetric/netdata",
+                "launch_order": 1,
+                "opts": [
+                    [
+                        "--cap-add=SYS_PTRACE",
+                        ""
+                    ],
+                    [
+                        "-v",
+                        "/proc:/host/proc:ro"
+                    ],
+                    [
+                        "-v",
+                        "/sys:/host/sys:ro"
+                    ]
+                ],
+                "ports": {
+                    "19999": {
+                        "description": "port used to access the webUI port",
+                        "host_default": 19999,
+                        "label": "webUI port",
+                        "ui": true
+                    }
+                }
+            }
+        },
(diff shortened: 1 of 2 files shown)


FroggyFlox:
I just added a commit to mount a volume to load custom configuration files, allowing the user to load custom plugins and customize the information displayed by Netdata.

phillxnet:
@FroggyFlox This looks like a really useful addition. Just tried it post override volume addition and all appears to be working as expected:

rock-on-test-netdata-sys-verview

I haven't however checked any of the config override capabilities myself but by default there is a tone of stuff that appears to work as expected.

Nice.

FroggyFlox:
Thanks for your enthusiastic feedback @phillxnet!
I tested the override 'feature' of this image and it does work as advertised: upon container start, all files placed in the mounted 'override' volume are copied back into `etc/netdata` which contains the original netdata configuration files.
Here are the [related lines](https://github.com/titpetric/netdata/blob/5d20c683685733c43bc40a2bd03189e846cc8cc3/releases/latest/scripts/run.sh#L109-L111) of the `run.sh` script triggered at every container's start:
```
for f in /etc/netdata/override/*; do
  cp -a $f /etc/netdata/
done
```

FroggyFlox:
Reference to this Rock-on was now added to the "root.json" file.

phillxnet:
Thanks @FroggyFlox much appreciated. Apologies for the unreasonable delay in merging this.

FroggyFlox:
No worries, @phillxnet, thanks a lot for merging this!

PR #161, opened 2018-09-23
Update Emby Rockon

### LINKED ISSUE #152: Update Emby Rock-on
Emby server recently released all its versions on DotNet and thus changed its docker image substantially. While the current Json Rock-on still works, updating it would bring a few new features.
https://hub.docker.com/r/emby/embyserver/

I'll try to spend time on this and am creating this issue to remind me and keep track it it.

FroggyFlox:
Fixes #152.

This updated Rockon includes a new "devices" object to allow users with compatible hardware to mount the corresponding device for hardware transcoding.
It also includes a new ENV variable `GIDLIST` to add system users with access to this device.

For backward compatibility, this keeps the `--net=host` option, but this also means that, as before, the port selection remains without actual use as the default emby ports can be used (8096 for HTTP, and 8920 for HTTPS). Would somebody know what needs this option?

### DIFF
--- a/embyserver.json
+++ b/embyserver.json
@@ -1,5 +1,5 @@
 {
-    "EmbyServer": {
+    "Emby server": {
         "containers": {
             "embyserver": {
                 "image": "emby/embyserver",
@@ -10,6 +10,12 @@
                         ""
                     ]
                 ],
+                "devices": {
+                    "VAAPI": {
+                        "description": "<u>Optional:</u> path to hardware transcoding device (/dev/dri/renderD128). Leave blank if not needed.",
+                        "label": "VAAPI device"
+                    }
+                },
                 "ports": {
                     "8096": {
                         "description": "Emby Server WebUI port. Suggested default: 8096",
@@ -17,6 +23,13 @@
                         "label": "WebUI port",
                         "protocol": "tcp",
                         "ui": true
+                    },
+                    "8920": {
+                        "description": "Emby Server HTTPS port. Suggested default: 8920",
+                        "host_default": 8920,
+                        "label": "HTTPS port",
+                        "protocol": "tcp",
+                        "ui": false
                     }
                 },
                 "volumes": {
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Thanks for these corrections / enhancements. I'm assuming all is OK here as you are more up on this code area's function than myself.

Noting however that we do have a Rock-on name change here, but I know you have tested this yourself locally.

Merging for now as we should be robust to this and if not then we can sort on the code side hopefully.

Thanks for the explicit mention of continued use of --net=host. Definitely something we need to re-visit across all our Rock-ons. Maybe some kind of 'red flag' by each in the UI if they use this. Again an area you are more familiar with.

FroggyFlox:
Thanks @phillxnet ,

> Noting however that we do have a Rock-on name change here, but I know you have tested this yourself locally.

As this was a Rock-on already existing remotely, that is actually something I actually wasn't sure how to test locally so hopefully that won't break things. As the entry in the `root.json` is the same that should be ok. My expected behavior would be that Rockstor detects that the existing entry in db is not in `root.json` anymore and thus delete it, before replacing it with the one in the `root.json`. I hope I didn't drop the ball on that one.

> Thanks for the explicit mention of continued use of --net=host. Definitely something we need to re-visit across all our Rock-ons. Maybe some kind of 'red flag' by each in the UI if they use this.
That's a good idea... I could integrate that and/or another way to change this in my current work on docker networks, but maybe that deserves a separate and dedicated issue and discussion.

phillxnet:
@FroggyFlox Opps sorry I forgot to enter my comment re this change in my current issue:

"missing root.json entries" #171

Have now committed the comment re my intended change in that file re this merge.

Let me know if I should hold off on that change in that issue?

FroggyFlox:
> Let me know if I should hold off on that change in that issue?

I believe it should be ok for users without this Rock-on installed, as well as for those with a current install of it, but if you prefer to stay on the conservative side of things, don't hesitate to revert the name change and keep it as before. That change was purely cosmetic after all... Sorry again for not being able to answer this for sure :-(

phillxnet:
@FroggyFlox Thanks. I think I'll go with the change as it's best if we maintain sync between the displayed name and what people may have to type in the 'emergency delete' script. Plus I think "Emby server" is more proper than the 2 words rolled together.

PR #167, opened 2018-10-13
Fix volume binding

### LINKED ISSUE #165: COPS Rockon incorrectly configures /movies
https://github.com/rockstor/rockon-registry/blob/master/cops.json#L21

^ Here we're setting up for books at the /movies path. COPS configuration in the image is looking for books at the /books path. This blocks COPS from running properly. 

Recommend updating the line to /books instead of /movies. I'm unable to try it / create a pull from my current environment.

FroggyFlox:
As discovered by @lgchris, the container requires a `/books` binding and not `/movies`. See the corresponding issue #165 for details.
Fixes #165 .

### DIFF
--- a/cops.json
+++ b/cops.json
@@ -18,7 +18,7 @@
                         "description": "Choose a Share for cops configuration. Eg: create a Share called cops-config for this purpose alone.",
                         "label": "Config Storage"
                     },
-                    "/movies": {
+                    "/books": {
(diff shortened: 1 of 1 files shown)


phillxnet:
@FroggyFlox Thanks for sorting out this fix and apologies to both you and @lgchris for the delay.

PR #168, opened 2018-11-13
Medusa Rockon

shadofall:
json file for the medusa PVR docker

### DIFF
--- a/medusa.json
+++ b/medusa.json
@@ -1,0 +1,53 @@
+{
+    "medusa": {
+        "containers": {
+            "medusa": {
+                "image": "linuxserver/medusa",
+                "launch_order": 1,
+                "ports": {
+                    "8081": {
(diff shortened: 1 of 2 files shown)


FroggyFlox:
Thanks for the PR @shadofall !

I tested it and I can attest it installs on 3.9,1-16, and its webUI is accessible. I couldn't test the actual functions of this Rock-ons but everything seems to work at fist.
It also shuts down properly and can be turned back on from the webUI without a problem.

@shadofall , would you be able to add this Rock-on to the root.json file as well so that it;ll be properly listed?

FroggyFlox:
Thanks for update @shadofall !

@schakrava, I believe this is ready for review.

phillxnet:
@shadofall Thanks for the Rock-on submission and apologies for taking so long.
And thanks to @FroggyFlox for reviewing yet another Rock-on, much appreciated and a great help.

PR #169, opened 2019-01-10
Plex official

coleberhorst:
Added a version of plex that uses the official image source and updates more frequently than current.

### DIFF
--- a/plex-official.json
+++ b/plex-official.json
@@ -31,13 +31,15 @@
                 "environment": {
                     "PLEX_UID": {
                         "description": "Enter a valid UID to run Plex with. It must have full permissions to all shares mapped in the previous step.",
-                        "label": "UID",
-                        "index": 1
+                        "label": "UID"
                     },
                     "PLEX_GID": {
                         "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all shares mapped in the previous step.",
-                        "label": "GID",
-                        "index": 2
+                        "label": "GID"
+                    },
+                    "PLEX_CLAIM": {
+                        "description": "The claim token for the server to obtain a real server token. If not provided, server will not be automatically logged in. If server is already logged in, this parameter is ignored. You can obtain a claim token to login your server to your plex account by visiting https://www.plex.tv/claim. Leave blank to sign in via plex.tv.",
+			            "label": "claim token (https://wwww.plex.tv/claim) (optional)"
                     }
                 }
             }
--- a/plex-official.json
+++ b/plex-official.json
@@ -6,7 +6,6 @@
                 "opts": [
                     [
                         "--net=host",
-                        "-e=TZ=$(timedatectl status|grep -oP 'zone:\\s+[\\w/]+'",
                         ""
                     ]
                 ],
--- a/plex-official.json
+++ b/plex-official.json
@@ -6,6 +6,7 @@
                 "opts": [
                     [
                         "--net=host",
+                        "-e=TZ=$(timedatectl status|grep -oP 'zone:\\s+[\\w/]+'",
                         ""
                     ]
                 ],
@@ -29,20 +30,15 @@
                     }
                 },
                 "environment": {
-                    "TZ": {
-                        "description": "Set the timezone inside the container. For example: Europe/London or America/Chicago. The complete list can be found here: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones",
-                        "label": "TIMEZONE",
-                        "index": 1
-                    },
                     "PLEX_UID": {
                         "description": "Enter a valid UID to run Plex with. It must have full permissions to all Shares mapped in the previous step.",
                         "label": "UID",
-                        "index": 2
+                        "index": 1
                     },
                     "PLEX_GID": {
                         "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
                         "label": "GID",
-                        "index": 3
+                        "index": 2
                     }
                 }
             }
--- a/plex-official.json
+++ b/plex-official.json
@@ -39,7 +39,7 @@
                     },
                     "PLEX_CLAIM": {
                         "description": "The claim token for the server to obtain a real server token. If not provided, server will not be automatically logged in. If server is already logged in, this parameter is ignored. You can obtain a claim token to login your server to your plex account by visiting https://www.plex.tv/claim. Leave blank to sign in via plex.tv.",
-			            "label": "claim token (https://wwww.plex.tv/claim) (optional)"
+			            "label": "token (https://wwww.plex.tv/claim)"
                     }
                 }
             }
--- a/plex-official.json
+++ b/plex-official.json
@@ -31,16 +31,12 @@
                 "environment": {
                     "PLEX_UID": {
                         "description": "Enter a valid UID to run Plex with. It must have full permissions to all shares mapped in the previous step.",
-                        "label": "UID"
+                        "label": "UID",
                     },
                     "PLEX_GID": {
                         "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all shares mapped in the previous step.",
                         "label": "GID"
                     },
-                    "PLEX_CLAIM": {
-                        "description": "The claim token for the server to obtain a real server token. If not provided, server will not be automatically logged in. If server is already logged in, this parameter is ignored. You can obtain a claim token to login your server to your plex account by visiting https://www.plex.tv/claim. Leave blank to sign in via plex.tv.",
-			            "label": "token (https://www.plex.tv/claim)"
-                    }
                 }
             }
         },
--- a/plex-official.json
+++ b/plex-official.json
@@ -24,7 +24,7 @@
                         "label": "Config Storage"
                     },
                     "/data": {
-                        "description": "Choose a Share for Plex content(your media). Eg: create a Share called plex-data for this purpose alone. You can also assign other media Shares on the system after installation.",
+                        "description": "Choose a share for Plex content (your media). Eg: create a share called plex-data for this purpose alone. You can also assign other media shares on the system after installation.",
                         "label": "Data Storage"
                     }
                 },
--- a/plex-official.json
+++ b/plex-official.json
@@ -29,15 +29,20 @@
                     }
                 },
                 "environment": {
+                    "TZ": {
+                        "description": "Set the timezone inside the container. For example: Europe/London or America/Chicago. The complete list can be found here: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones",
+                        "label": "TIMEZONE",
+                        "index": 1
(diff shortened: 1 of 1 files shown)


phillxnet:
@coleberhorst Thanks for the submission.
I'd like to request a change however:
You have manual timezone entry but we can set this auto style via the build in mapping.
See the following pull request where this was also added inadvertently:
"Duckdns" #138
The docker image should pick on on the machines local time. This will also simplify setup and avoid users having to look up time zone codes.

phillxnet:
Linking to your forum thread relating to this pull request for context:

"Plex from official image"
https://forum.rockstor.com/t/plex-from-official-image/5676

where @FroggyFlox and forum user Haioken address some issues with this pr.

Thanks again for stepping up to supply an official Plex rock-on.

FroggyFlox:
Thanks for the update @coleberhorst,

> Passing the system timezone to the -e switch ended up working and I updated my pull request.

Have you tried without specifying the TZ at all? As mentioned in the corresponding forum post, Rockstor automatically binds the system's localtime to this end so you may not need to.
See the corresponding code: [https://github.com/rockstor/rockstor-core/blob/33917f112011d13e702633f525f2afdc0de78238/src/rockstor/storageadmin/views/rockon_helpers.py#L186-L187](https://github.com/rockstor/rockstor-core/blob/33917f112011d13e702633f525f2afdc0de78238/src/rockstor/storageadmin/views/rockon_helpers.py#L186-L187)

coleberhorst:
ok it does seem to work without the timezone binding. i tried a version without timezone, but maybe some other problem prevented it from loading. Works now, I think it is ready for merge and/or someone else to test it.

phillxnet:
@coleberhorst Thanks for you effort on this one and sorting that timezone bit out.

Lets get a review and if all is well this can go it.

plex-official.json
```diff
@@ -0,0 +1,54 @@
...
+                    }
+                },
+                "environment": {
+                    "PLEX_UID": {
+                        "description": "Enter a valid UID to run Plex with. It must have full permissions to all Shares mapped in the previous step.",
+                        "label": "UID",
+                        "index": 1
+                    },
+                    "PLEX_GID": {
+                        "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
```
FroggyFlox: Comply with Rockstor standard format ([rockstor/rockstor-core#1914](https://github.com/rockstor/rockstor-core/pull/1914)):
```suggestion
                        "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all shares mapped in the previous step.",
```

plex-official.json
```diff
@@ -0,0 +1,54 @@
...
+                        "label": "GID",
+                        "index": 2
+                    }
+                }
+            }
+        },
+        "description": "Official Plex Media Server Docker Repo",
```
FroggyFlox: Comply with Rockstor standard format ([rockstor/rockstor-core#1914](https://github.com/rockstor/rockstor-core/pull/1914)):
```suggestion
        "description": "Official Plex Media Server docker repository",
```

plex-official.json
```diff
@@ -0,0 +1,54 @@
...
+        "more_info": "<h4>Adding more media to Plex.</h4><p>You can add more Shares(with media) to Plex from the settings wizard of this Rock-on. Then, from Plex WebUI, you can update and re-index your library.</p>",
```
FroggyFlox: Comply with Rockstor standard format ([rockstor/rockstor-core#1914](https://github.com/rockstor/rockstor-core/pull/1914)):
```suggestion
        "more_info": "<h4>Adding more media to Plex.</h4><p>You can add more shares (with media) to Plex from the settings wizard of this Rock-on. Then, from Plex WebUI, you can update and re-index your library.</p>",
```

phillxnet:
@FroggyFlox re accessing the Plex rock-on Web-UI; we may have had something similar before with our existing Plex container. See:
"add plex ui access tip on rock-on write-up"
https://github.com/rockstor/rockstor-doc/issues/172
which was a suggested doc work around.
May not be relevant here though.

FroggyFlox:
Thanks for the link, @phillxnet!

Unfortunately, I get a browser error ("the connection was reset" if I remember correctly) and cannot even connect the Plex "Sign-in" page. I can if I follow the workaround documented on docker hub (use ssh tunneling to link localhost on a user machine to the server's ip).
Interestingly I do not have this problem with the existing Linuxserver.io Plex rock-on. This may be limited to myself, though, but we would need confirmation by somebody else.

coleberhorst:
Ok I tried to test by deleting my plex-config share and replacing the json with a freshly downloaded one from this pull request. It worked well and had to do the usual server fresh setup, but I was still signed into plex (maybe plex pulls shared cookies from plex.tv signin idk. So not a conclusive test.

Tried again by deleting plex-config, rockons share (in case it had any credentials or something), and signed out of plex on my browser and even opened it in icognito mode. I was redirected to a sign on screen from the normal app.plex.tv. So I cannot verify any issue requiring ssh to access the web gui on Chrome- Windows 10.

I did experience a weird bug where rockstor only presented me the option to choose a share for /data and not for /config. I pressed update and it requested both. I'm not sure if this is fixable in the .json or just one of the quite a few rockon webui bugs I have experienced.

coleberhorst:
Oh some plex installs, I have seen rockstor require several minutes to load plex because the install is quite large compared to other images. Maybe that's it?

magicalyak:
Don't you want to add this?
PLEX_CLAIM The claim token for the server to obtain a real server token. If not provided, server will not be automatically logged in. If server is already logged in, this parameter is ignored. You can obtain a claim token to login your server to your plex account by visiting https://www.plex.tv/claim

magicalyak:
I tested the rockon and it works but adding the CLAIM would help.  I added it and it saves the setup step.  Load was quick for this container also.

`
.... (under environment)
                    },
		    "PLEX_CLAIM": {
			"description": "The claim token for the server to obtain a real server token. If not provided, server will not be automatically logged in. If server is already logged in, this parameter is ignored. You can obtain a claim token to login your server to your plex account by visiting https://www.plex.tv/claim",
			"label": "claim token (https://wwww.plex.tv/claim)",
			"index": 3
		    }

`

coleberhorst:
@phillxnet is there a way to make this claim token optional? 

I'm fine with including it, but I'm unsure whether the rockon "philosophy" is require the least amount of input from the user. My personal opinion is the plex.tv redirect page is very clean; however, if you like the claim token or claim token option idea, I'll add it.

magicalyak:
If you leave it blank it won’t set. I guess we just need to say that better in the input if it’s not clear.

coleberhorst:
Ok when testing the claim code, leaving it blank causes rockstor to complain that it is required. So I'll just wait for a response to my question about how to make environment vars optional?

If there isn't a way to make optional environment vars, I'm tempted to revert because adding a box and link to the seems awkward when clicking the Plex-official UI button brings you to the sign in page. I guess it's more of a rockon philosophy question at that point: is it better to have less required inputs or not?

magicalyak:
It depends lol!  I know if I used the rockon I'd want the auto-login and would input the tag (it's only during install and a blank box is accepted).  I can't do that if the box isn't there.  However, the linuxserver.io image doesn't have it and I seem to use a cookie to just login anyway.  I'm fine either way honestly.  We can also open an issue and PR this if someone wants it later but I thought it would be good to have now.

magicalyak:
Also is this ready to test?

coleberhorst:
Yessir!

coleberhorst:
Thanks for testing everyone! My only reservation is that the current version forces the user to enter a claim code (leaving blank doesn't work). If that's ok with you guys, I'll just leave it in. 

I'd need to remove the language about leaving it blank from the description before it gets merged. Unless there's a way to allow it to be blank / optional? Other option is removing it. Need some feedback from others on this.

magicalyak:
If it won’t work blank remove that part and let’s go without it.  Sorry to make this difficult.

coleberhorst:
ok reverted. ready for merge.

plex-official.json
```diff
@@ -0,0 +1,52 @@
...
+                    },
+                    "/data": {
+                        "description": "Choose a share for Plex content (your media). Eg: create a share called plex-data for this purpose alone. You can also assign other media shares on the system after installation.",
+                        "label": "Data Storage"
+                    }
+                },
+                "environment": {
+                    "PLEX_UID": {
+                        "description": "Enter a valid UID to run Plex with. It must have full permissions to all shares mapped in the previous step.",
+                        "label": "UID",
```
phillxnet:
```suggestion
                        "label": "UID"
```

phillxnet: I think we just had a rouge comma added here.

coleberhorst:
awesome thanks for catching those. My revert wasn't as clean as I thought.

phillxnet:
@coleberhorst Well done on persevering with this. It's now undergone a goodly number of changes and had a number of reviews so can we have a final test of what this pr represents now and if it works as expected then lets get it in.

coleberhorst:
I confirmed the latest json works.

Hooverdan96:
The only challenge I see with the official image is that it doesn't offer the additional architectures like the linuxserver does (in reference to this discussion issue by @mcbridematt : https://github.com/rockstor/rockstor-core/issues/2191

And finally, one more question for @coleberhorst , on the docker hub there is this blob about the tag (https://hub.docker.com/r/plexinc/pms-docker - section on Tags):

> In addition to the standard version and latest tags, two other tags exist: _plexpass_ and _public_. These two images behave differently than your typical containers. These two images do not have any Plex Media Server binary installed. Instead, when these containers are run, they will perform an update check and fetch the latest version, install it, and then continue execution. They also run the update check whenever the container is restarted. To update the version in the container, simply stop the container and start container again when you have a network connection. The startup script will automatically fetch the appropriate version and install it before starting the Plex Media Server.

Since in the json you don't specify a version (i.e. translates into "latest") - does that mean that the version does not get updated unless the container is forcibly removed and then newly pulled (as opposed to a stop/restart triggering an update/upgrade)? If that's the case, we might have to mention that in this RockOn

phillxnet:
@FroggyFlox @Hooverdan96 and @coleberhorst I'm thinking now this pr has reached the no-activity threshold and should be closed. My apologies to all concerned but I think we should now shy away from multiple Rock-ons essentially offering a near identical instance of a project, such as Plex in this case.

Such customisation is always still available to folks via the locally hosting Rock-on definitions but from now on I think we should really try for only quite different instances of one project and then only if there is real value in having the different/parallel versions.

Also we have now gotten a lot of house work to do so having to be a little more clean cut with getting rid of stuff that has not received appropriate attention.

Obviously anyone concerned should chime in if they fee this is an overstep but we must now reduce this repo to something more fluid/flowing and reducing it's backlog is a part of who I would like to approach that.

Hooverdan96:
Agreed. I think we can close this one for now.

PR #170, opened 2019-01-11
Add Bitwarden-rs alpine

guppy0130:
Rockon for [this unofficial bitwarden docker image](https://github.com/dani-garcia/bitwarden_rs)

Written in rust, uses less resources than the official image, and the image runs on Alpine, so it's even *smaller*.

Some setup required:
* TLS/SSL certs + key: full chain cert must be at `/ssl/certs.pem` and key at `/ssl/key.pem` (I made this share first, turned it into an SMB share, placed the files in, then pointed the Rockon at the share, but as long as the files exist in the right spot you're good to go)

Container port 80 mapped out to whatever, default doesn't really matter so feel free to change it to whatever doesn't clash with existing systems.

Confirmed working on Rockstor 3.9.1-16, Bitwarden 2.8.0 (docker tag 1.5.0-alpine or something close to that), Chrome extension 1.37.0, and Android app 1.21.0 (but should work with all versions)

### DIFF
--- a/bitwarden-rs-alpine.json
+++ b/bitwarden-rs-alpine.json
@@ -7,6 +7,7 @@
             "bitwarden-rs": {
                 "image": "mprasil/bitwarden",
                 "tag": "alpine",
+                "launch_order": 1,
                 "ports": {
                     "80": {
                         "description": "Admin interface",
--- a/bitwarden-rs-alpine.json
+++ b/bitwarden-rs-alpine.json
@@ -9,7 +9,7 @@
                 "tag": "alpine",
                 "launch_order": 1,
                 "ports": {
-                    "80": {
+                    "443": {
                         "description": "Admin interface",
                         "label": "Admin interface",
                         "host_default": 80,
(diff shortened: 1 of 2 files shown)


bitwarden-rs-alpine.json
```diff
@@ -0,0 +1,36 @@
+{
+    "Bitwarden-rs": {
+        "description": "Bitwarden-rs on Alpine",
```
FroggyFlox: I would develop this a little bit more, so that users unfamiliar with Bitwarden know what it is used for. A single sentence should suffice.

bitwarden-rs-alpine.json
```diff
@@ -0,0 +1,36 @@
...
+        "version": "0.0.1",
+        "website": "https://github.com/dani-garcia/bitwarden_rs",
+        "containers": {
```
FroggyFlox: You could use the `"more_info"` object to add more details on how to setup certs (or a link). You can embed html tags to make it more legible. See the [ZoneMinder Rock-on](https://github.com/rockstor/rockon-registry/blob/bd97e1a93ee28e6a0bf75ad22c5836c02ba0eba8/zoneminder-1.30.json#L37) by @magicalyak  for an example.

phillxnet:
@guppy0130 Thanks from me also for another Rock-on submission. Much appreciated.
And thanks to @FroggyFlox for the review and valuable suggestions.
Lets pop this one in now it has @FroggyFlox's suggestions and see how it goes.

PR #174, opened 2019-02-03
added pocketmine with bedrock edition update

magicalyak:
Added minecraft pocket edition server to rockons.  This uses the new bedrock alpha server that doesn't need constant babysitting or restarts.  Although its alpha its more stable than the last release on PHP.


magicalyak:
Will do, definitely not ready yet.  We don't have a default for environment items do we (I mean you can set it but it won't fill it in).

phillxnet:
@magicalyak Don't think so no. Although @FroggyFlox is more up on the Rock-on system as has been working in that area quite a bit of late.

magicalyak:
I think I need to modify my docker hub image and hard code the ZIP download which is frustrating.  I think the error rate on having people type it in is too hard.

phillxnet:
@magicalyak Thanks for taking the time to prepare this again by the way. Much appreciated and will be a nice addition.

phillxnet:
@magicalyak Sound like the way to go actually. More self contained that way also. Plus from Rockstor's perspective it's an easier setup.

FroggyFlox:
> We don't have a default for environment items do we (I mean you can set it but it won't fill it in).

Hi @magicalyak, to the best of my knowledge, NO, we do not. It is, however, a good idea and should not require too much to implement I believe. That would deserve its own issue on the [rockstore-core repo](https://github.com/rockstor/rockstor-core). Would you mind opening one? If not I'll take care of it later once I can.

With regards to hardcoding the URL, I'm afraid it is the easiest way for the user, indeed. How would different versions of the server be managed, however? If you end up having to make your own image, would it be imaginable to have a `-e $VERSION` entry in the JSON and use it to direct the `entrypoint.sh` to download the requested ZIP? We would thus only asks the user to input its version of choice (such as 1.13, 1.12, etc..., all listed in the "help" icon). I'm not sure this is easily implementable, however, so I'm referring to your expertise on that one.

magicalyak:
Still working on this and I wonder if it's me.  I can get the container to run perfect on CLI but this rockon is giving me fits.  I'm testing on a clean system because I'm thinking it's my system at this point

magicalyak:
Wow, been going nuts trying to figure out why this won't work and it's me.  I spun up a new system and this works perfect.  It needs -it for the console to work, otherwise you can't issue commands to it (such as list and op user).  It's passed in the opts section and works great.
@phillxnet this is ready to merge!

phillxnet:
@magicalyak Well done on persevering, and thanks for sharing your findings / explanation re the "-it" flag; I see this came up previously in your original pr for this Rock-on (linked below).

Merging as per your review. I'll add the root.json entry shortly after merge.

Thanks also to @FroggyFlox for technical and usability input on this one, interesting idea re enhancements also. But lets get this one in as is for now especially given it's log tail in connection with the original pr #49 (2016 !!) where we also owe thanks to @schakrava for initial review and early technical guidance.

PR #181, opened 2019-02-26
Added crashplanpro.json

wiza:
Rockon for crashplan pro.

### DIFF
--- a/crashplanpro.json
+++ b/crashplanpro.json
@@ -1,0 +1,39 @@
+{
+	"Crashplan-Pro": {
+        "containers": {
+            "crashplan-pro": {
+                "image": "jlesage/crashplan-pro",
+                "launch_order": 1,
+		"ports": {
+                    "5800": {
+                        "description": "WebUI port.",
+                        "host_default": 5800,
+                        "label": "WebUI port",
+                        "ui": true
+                    },
+                    "5900": {
+                        "description": "VNC port.",
+                        "host_default": 5900,
+                        "label": "VNC port"
+                    }
+                },
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a share for the crashplan configuration & log files.",
+                        "label": "/config"
+                    }
+		},
+		"opts": [ 
+			[
+				"-v", 
+				"/mnt2:/storage"
+			]
+		]
+            }
+        },
+        "description": "Crashplan Pro rockon, container from jlesage/crashplan-pro",
+	"website": "https://hub.docker.com/r/jlesage/crashplan-pro/",
+        "volume_add_support": true,
+        "version": "beta"
+    }
+}


phillxnet:
@wiza Thanks for the Rock-on submission. Much appreciated.
As a suggestion would you mind making the Description fields a little more beginner friendly: i.e. take a look at our existing non pro version of crashplan where the descriptions/user facing text are more in line with where we are heading.
https://github.com/rockstor/rockon-registry/blob/master/crashplan.json

Also note that we are moving to have the website link to the upstream project and incorporate into the general description the docker hub image used. An example of this is in the following json:
https://github.com/rockstor/rockon-registry/blob/master/teamspeak3.json

Thanks again. I'll mark this as needing a review for now.

wiza:
Will do once I have the time. BTW. crashplan has no personal plan anymore.

phillxnet:
@wiza 
> BTW. crashplan has no personal plan anymore.
OK, I didn't realise that. So maybe it would be best to edit the existing crashplan json if it uses a now defunct docker image.
Thanks again for sorting this Rock-on out by the way. Let us know when it's ready for review.

wiza:
I just stopped using crashplan, they changed terms so that it isn't really useful (crashplan which does not backup configuration files, time machine backups etc is not actually a part of a crashplan).

Noggin01:
As @wiza wiza commented, Crashplan Home (the only current Crashplan Rock-On) has been discontinued.  It should either be removed from the Crashplan repo or renamed to indicate that it is not recommended for installation as it probably won't do anything at all.

However, I've installed Crashplan PRO using the json file provided by wiza and am currently backing up the files I wish to have backed up.

@philnext, you mentioned that it might be best to edit the existing crashplan json.  If you mean to edit that json to install CrashPlan PRO, please ensure that the name of the Rockon is also changed appropriately so as to not confuse anyone that is familiar with old CrashPlan product line.

@wiza, this is not likely the correct place to have this discussion, but `they changed terms so that it isn't really useful` I think is unfair.  They changed the terms so that you can't back up locations such as "Program Files" and virtual machines.  I don't think that this makes CrashPlan PRO backups unuseful.  Rockstor is a NAS and I think it is well suited for people learning to administer this type of system.  Linux doesn't have a whole lot of options for all-in-one online backup solutions and I think this makes CrashPlan PRO a great option _especially_ since you got this Rockon going!

wiza:
Feel free to use it :) I meant it isn't useful to me as a crashplan (meaning a plan to get everything up and running fast after a crash). I did myself a another NAS running opensuse on a terramaster box with btrbk and crypted disks and run it on another country for redundancy :D

I did have these opts added to the rockon until I stopped using it to get it to run as root to have access to _everything_:

		"opts": [ 
			[
				"-v", 
				"/mnt2:/storage:ro"
			],
			[
				"-e",
				"USER_ID=0"
			],
			[
				"-e",
				"GROUP_ID=0"
			]
		]

wiza:
@phillxnet also, pro is a different product from home so probably easier to remove old one and if someone takes up pro maintenance, it should probably be named as pro. Old one doesn't work anyway.

PR #186, opened 2019-04-15
add jdownloader2 rock-on

DarkYamik:
As requested on my post I created the pull request to add Jdownloader rockon :)

### DIFF
--- a/Jdownloader2.json
+++ b/Jdownloader2.json
@@ -50,7 +50,7 @@
 			"slug": ""
 		},
 		"volume_add_support": true,
-		"website": "https://github.com/jlesage/docker-jdownloader-2",
+                "website": "http://jdownloader.org/",
 		"version": "1.0"
 	}
 }
--- a/Jdownloader2.json
+++ b/Jdownloader2.json
@@ -1,0 +1,56 @@
+{
+	"Jdownloader 2": {
+		"containers": {
+			"jdownloader2": {
+				"image": "jlesage/jdownloader-2",
+				"launch_order": 1,
+				"ports": {
+					"5800": {
+						"description": "WebUI Port",
+						"host_default": 5800,
+						"label": "Data Port",
+						"protocol": "tcp",
+						"ui": true
+					},
+					"3129": {
+						"description": "MyJdownloader Port",
+						"host_default": 3129,
+						"label": "MyJdownloader Port",
+						"protocol": "tcp"
+					}
+				},
+				"volumes": {
+					"/config": {
+						"description": "Choose a Share for configuration. Eg: create a Share called jdownloader-config for this purpose alone.",
+						"label": "Config Storage",
+						"min_size": 1073741824
+					},
+					"/output": {
+						"description": "Share where all the downloads will be",
+						"label": "Data Storage"
+					}
+				},
+				"environment": {
+					"PUID": {
+						"description": "UID to run Jdownloader ",
+						"label": "UID",
+						"index": 1
+					},
+					"PGID": {
+						"description": "Enter a valid GID to use along with the same UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
+						"label": "GID",
+						"index": 2
+					}
+				}
+			}
+		},
+		"description": "JDownloader 2 Rock-On using jlesage docker",
+		"ui": {
+			"https": false,
+			"slug": ""
+		},
+		"volume_add_support": true,
+		"website": "https://github.com/jlesage/docker-jdownloader-2",
+		"version": "1.0"
+	}
+}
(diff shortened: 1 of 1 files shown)


phillxnet:
@DarkYamik Thanks for the pull request. Much appreciated.

DarkYamik:
Hi @FroggyFlox  and thank you for reviewing this Rock-on, much appreciated!

If the `SECURE_CONNECTION` option is as easy as apply it to the .json and it works properly as you confirmed, I see no problem on enabling it by default.

If you could add the suggestion on how you enabled it, i'll commit it as soon as I'm able to!

Thank you for your review!

FroggyFlox:
Hi @DarkYamik,

Thanks a lot for integrating my suggestions, I apologize for not getting back with you earlier; the past few days have been very busy for me and I couldn't find time to properly test the pros & cons of using the `SECURE_CONNECTION`. I actually am undecided on the best way to deal with it, and I see the following different possibilities:
1. We force SECURE_CONNECTION enabled in the `"options"` object, which means the user does not have the choice. We can thus set the ui link to use `"https": true` and everything will be seamless to the user. Nevertheless, the fact that the author of the docker image set this option to disabled by default makes me thing there was a good reason for it. In this case, forcing it to be enabled may cause issue in some (fringe?) cases.
2. To avoid forcing it to users, we can use the `:environment"` object and let the user decide.. We have, however, some incompatibilities that emerge with the `"https"` parameter. Indeed, in this case we have;
2a. if `SECURE_CONNECTION` is enabled and `"https"` set to true: everything works well and is seamless.
2b. if `SECURE_CONNECTION` is enabled and `"https"` set to false: the browser will throw the "Insecure connection warning". Although it will let user bypass this warning, it isn't seamless at all.
2c. if `SECURE_CONNECTION` is disabled and `"https"` set to true: browser error.
2d. if `SECURE_CONNECTION` is enabled and `"https"` set to false: no problem.
3. We let it as it, without option to enable `SECURE_CONNECTION`.

@DarkYamik, @phillxnet, any input on the most elegant way to deal with this?
My preference would go to option 3, especially given the fact that a more secure connection can be achieved by placing the container behind a reverse-proxy, for instance. This, and given the fact that the image's default is to let this option disabled...

Jdownloader2.json
```diff
@@ -0,0 +1,56 @@
...
+					},
+					"PGID": {
+						"description": "Enter a valid GID to use along with the same UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
+						"label": "GID",
+						"index": 2
+					}
+				}
+			}
+		},
+              "description": "JDownloader 2 Rock-On using jlesage docker.<p>Based on the following docker image: <a href='https://github.com/jlesage/docker-jdownloader-2' target='_blank'>https://github.com/jlesage/docker-jdownloader-2</a>.",
```
FroggyFlox: Maybe we can use the description of the project as provided by the authors? The following below is the "official" description found on Jdownloader doc page:

```suggestion
              "description": "<p>JDownloader is a free, open-source download management tool with a huge community of developers that makes downloading as easy and fast as it should be. Users can start, stop or pause downloads, set bandwith limitations, auto-extract archives and much more. It's an easy-to-extend framework that can save hours of your valuable time every day!</p><p>Based on the following docker image: <a href='https://github.com/jlesage/docker-jdownloader-2' target='_blank'>https://github.com/jlesage/docker-jdownloader-2</a>.",
```

phillxnet:
@FroggyFlox Re:
> @DarkYamik, @phillxnet, any input on the most elegant way to deal with this?
> My preference would go to option 3, especially given the fact that a more secure connection can be achieved by placing the container behind a reverse-proxy, for instance. This, and given the fact that the image's default is to let this option disabled...

Agreed, especially given our mid term plans to provide a reverse proxy facility.
And thanks from me as well for reviewing this. Much appreciated.

@DarkYamik I know it's been ages but if you fancy committing @FroggyFlox's changes, ie via "Commit suggestion" we can get this one done and dusted. Thanks for the contribution and your patience.

DarkYamik:
Hello! @FroggyFlox @phillxnet 
First of all sorry for not commiting the pull request before, i had some work related issues and I totally forgot about this.
I added commited @FroggyFlox changes 

Thank you both for your time and patience!

phillxnet:
@DarkYamik Thanks for applying the suggestions.
@FroggyFlox Thanks again for helping to get this one in.

FroggyFlox:
Thanks @DarkYamik for the changes!
@phillxnet, thanks for manually adding it to `root.json`. Sorry I forgot to request that change as well... I always tend to forgot that one.

phillxnet:
@FroggyFlox No worries, ideally we would have that automated some how, That way contributors have only the one file to think about.
JDownloaded Rock-on should be published shortly.

PR #189, opened 2019-05-17
add zoneminder latest rock-on

magicalyak:
Zoneminder 1.32.3
needed to use a new rockon because the others aren't directly upgradeable.  I also had to host the docker image on my repository but this should help those wanting to use the latest zoneminder.
@phillxnet @schakrava merge when you can, thanks!


phillxnet:
@magicalyak Thanks for stepping up to this one, marking as needs review.

zoneminder-latest.json
```diff
@@ -0,0 +1,80 @@
...
+        "more_info": "</h4>Tips and Setup Instructions:</h4></p>        </p>        This container includes mysql, no need for a separate mysql/mariadb container</p>            All settings and library files are stored outside of the container and they are preserved when this docker is updated or re-installed (change the variable /path/to/config in the run command to a location of your choice)</p>                    Important:</p>                    </p>                    The web gui will be available at https://serverip:port/zm</p>                    On first start, open zoneminder settings, go to the paths tab and enter the following for PATH_ZMS: /zm/cgi-bin/nph-zms</p>                        The default timezone for php is set as America/New_York if you would like to change it, edit the php.ini in the config folder. Here's a list of available timezone options: http://php.net/manual/en/timezones.php",
+        "ui": {
+            "https": true,
+            "slug": "zm"
+        },
+        "volume_add_support": true,
+        "version": "1.32",
+        "website": "https://github.com/dlandon/zoneminder"
```
FroggyFlox: Link to project itself rather than docker image:

```suggestion
        "website": "https://zoneminder.com"
```

zoneminder-latest.json
```diff
@@ -0,0 +1,80 @@
...
+                    },
+                    "/var/cache/zoneminder": {
+                        "description": "Choose a Share for the cache.",
+                        "label": "ZoneMinder Cache",
+                        "min_size": "1073741824"
+                    }
+                }
+            }
+        },
+        "description": "ZoneMinder: Free, open-source software to control IP, USB and Analog (CCTV) cameras (v1.32) - please note this runs as privileged in docker (to set shm to a higher amount)",
```
FroggyFlox: Add link to docker image:

```suggestion
        "description": "ZoneMinder: Free, open-source software to control IP, USB and Analog (CCTV) cameras (v1.32) - please note this runs as privileged in docker (to set shm to a higher amount).<p>Based on the following docker image: <a href='https://github.com/dlandon/zoneminder' target='_blank'>https://github.com/dlandon/zoneminder</a></p>.",
```

phillxnet:
@magicalyak If you fancy having a look over @FroggyFlox suggestions and resolving / committing I'm hoping we will be their on this one also.
Cheers.

magicalyak:
I'm wondering if I should retest with this one? https://github.com/ZoneMinder/zmdockerfiles, there is no user or group specification though but it's the official one....
@phillxnet @FroggyFlox thoughts?

phillxnet:
@magicalyak If you could first commit @FroggyFlox suggestions that would be good as they were based on a review at the time so would be good to not have them go to waste, if you agree with them that is. And of course their is no reason why we shouldn't have multiple rock-ons if their is sense in doing so of course. That way we could get this one in, as it is nice to have user / group, and pop the official one in as well under an 'official' variant.

magicalyak:
Should be good now @phillxnet

phillxnet:
@magicalyak Thanks for seeing this one through and well done on your preparation and maintenance of this docker config.
Merging as discussed but I'm thinking we are now at 3 variants of ZoneMinder and given 2 are from yourself, and with the potential addition of a 4th official one, I'm again tending towards some pruning. But I do appreciate that their have been some major changes in these versions. Feel free to do a pr to remove your prior ZoneMinder-1.30 Rock-on if this one can be considered a drop-in replacement.

phillxnet:
@magicalyak This Rock-on is now published. Thanks.

I think it would nice at some point if we could, as a matter of course, include example names for each share as we do in a number of other Rock-ons, and a little more info about what 'cache' means in the context of this Rock-on. I.e. is that where all their recorded footage will end up for example? But it's in now and we can always tweak.
Thanks again for your numerous contributions and do have a think about redundancy as I think we are approaching what should be peak Rock-on kipple :)

PR #193, opened 2019-09-07
Change deluge to --net=host to allow remote connec

Psykar:
As per https://github.com/linuxserver/docker-deluge/blob/master/Dockerfile the port 58946 should be exposed.
And for remote connections while seeding to actually be possible, the network mode needs to be 'host' (see https://github.com/linuxserver/docker-deluge#parameters )

### DIFF
--- a/deluge.json
+++ b/deluge.json
@@ -18,8 +18,15 @@
                         "label": "Daemon port",
                         "protocol": "tcp",
                         "ui": true
+                    },
+                    "58946": {
+                        "description": "Deluge Remote connection port. Suggested default: 58946",
+                        "host_default": 58946,
+                        "label": "Remote connection port",
+                        "ui": true
                     }
                 },
+                "opts": ["--net", "host"],
                 "volumes": {
                     "/config": {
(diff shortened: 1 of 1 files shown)


phillxnet:
@Psykar Apologies for not response here. We are trying to avoid the heavy hammer of --net=host but I'm unsure of how we get around this in such a circumstance. We have just had a related issue opened by @Arderos, linking back for context:

"Deluge json has no incoming port defined" #205

phillxnet:
@Psykar re the ongoing discussion in the recently linked issue I'm inclined to close this pull request given the general consensus that we want to avoid --net=host use in Rock-ons as much as possible. In this case it may not be avoidable and in which case I'd rather drop the Rock-on all together. But I'm hoping we can arrive at a doc entry config walk through solution which means we can keep this Rock-on without us having to adopt yet another Rock-on with --net=host capabilities.

Thanks for helping to bring attention to this currently malfunctioning Rock-on and detailing it's currently missing port entry. We will have this pr (in closed state) as reference for future pull request once we sort out the direction we are to take with it. We can continue that discussion in the references thread.

Thanks again for the pull request and supplied, and referenced, details.

PR #202, opened 2019-10-24
Add YouTrack Official rock-on

FroggyFlox:
Fixes #201.

### General information on project
This pull request proposes to add a new rock-on for the following project:
- name: YouTrack
- website: https://www.jetbrains.com/youtrack
- description: Issue tracker designed for development teams. 

### Information on docker image
- docker image: https://hub.docker.com/r/jetbrains/youtrack
- is an official docker image available for this project?: yes. Note that the version needs to be hardcoded using the image tag. This PR uses the latest available: 2019.2.58831.

### Checklist
- [x] Passes [JSONlint](https://jsonlint.com) validation
- [x] Entry added to `root.json` in _alphabetical_ order (for new rock-on only)
- [x] `"description"` object lists and links to the docker image used
- [x] `"description"` object provides information on the image's particularities (advantage over another existing rock-on for the same project, for instance)
- [x] `"website"` object links to project's main website

@phillxnet, as this rock-on needs specific pre-install setup related to ownership of shares, I was thinking of documenting these first a write-up in Rockstor documentation to provide step-by-step instructions (see https://github.com/rockstor/rockstor-doc/issues/211) and linking to it in this rock-on's description. As a result, this PR would be dependent on https://github.com/rockstor/rockstor-doc/issues/211, which explains why I'm leaving it in "_Draft_" state for now. Thanks for letting me know what you think of the idea.

PR #203, opened 2019-10-26
Link to the correct section of the Rockstor documentation

FroggyFlox:
Link to the correct section of the Rockstor documentation for instructions on meeting the pre-install requirements for the YouTrack rock-on, as discussed in #202.

Note that this PR also deletes the now deprecated older youtrack.json rock-on. Its entry was removed from `root.json` in #202 but the actual `youtrack.json` file was left untouched.

@phillxnet, I hope I didn't miss anything again this time.


phillxnet:
@FroggyFlox That's great thanks.
Looks to be sorted now so merging and final test of resulting repo before publishing.

No worries if we have missed something as we can just patch again.

PR #226, opened 2020-06-08
Minecraft Bedrock Edition Server 

ROR89:
To cover the universe of minecraft this time in its version for console and native windows 10.
great thanks to colleagues: 
Hooverdan
Flox
phillxnet
 who without them never been possible this small project


phillxnet:
@ROR89 Well done and thanks for getting this sorted. We just need a review, which is likely to suggests some minor changes, and we can hopefully then get this merged and published.

But one simple suggesting is that we like to give folks a link, within the description, to the docker image used. It's a bit scrappy how we do this currently but does present a better user interface. Take a look at the other more modern Rock-ons and you will see how it's done.

If you look at the following template (by @FroggyFlox ) for new Rock-on pull requests:

https://github.com/rockstor/rockon-registry/blob/master/.github/pull_request_template.md

you will see a checklist that would be good to include in your original post (by editing it's text). Once saved you can then tick them off as you get them sorted.

Hope that helps and again, thanks for taking the effort to get this far.

magicalyak:
The current pocketmine should be working, what's the difference between this and that one?  They both are bedrock.

phillxnet:
As per @magicalyak outstanding question (2 years ago) I'll close this pr as now redundant. Do please re-open or better still re-submit ontop of our current master if there is interest.

PR #237, opened 2020-09-03
Update netdata rock-on. Fixes #236

### LINKED ISSUE #236: Netdata causing undesired AppArmor behavior
Upon installing Rockstor 4.0.1.0, enabling and configuring Rock-On service and installing the Netdata rockon the following appears cyclically on the console and in dmesg:

```
[404887.813445] audit: backlog limit exceeded
[404887.813505] audit: audit_backlog=65 > audit_backlog_limit=64
[404887.815997] audit: type=1400 audit(1599054448.661:817848324): apparmor="DENIED" operation="ptrace" profile="docker-default" pid=22193 comm="apps.plugin" requested_mask="read" denied_mask
="read" peer="unconfined"
[404887.817281] audit: audit_lost=12218207 audit_rate_limit=0 audit_backlog_limit=64
```

The rate is about one cycle od messages per second 

The installation of Rockstor was performed by an image (OpenSUSE 15.2) created by kiwi-ng.
Upon check AppArmor and auditd are either disabled (apparmor) or not present at all (auditd) on the system.

FroggyFlox:
Fixes #236.
@phillxnet , ready for review.

As detailed in issue #236, we need to update our current Netdata rock-on to fix a few problems:
- log spam by auditd due to misconfiguration of the container with regards to Apparmor
- out-of-date version of Netdata due to not using the official image

This PR thus proposes to:
- correctly use the correct apparmor profile (unconfined)
- switch to the official netdata/netdata image as it is kept up-to-date, is multi-arch, and allows for better customizations. The only drawback is the loss of the configuration override volume that was bound in our current rock-on, but I believe it isn't a big problem anyway.
- update links to website (general project and documentation)

### Checklist

### DIFF
--- a/netdata-official.json
+++ b/netdata-official.json
@@ -1,5 +1,5 @@
 {
-    "Netdata": {
+    "Netdata (official)": {
         "containers": {
             "netdata": {
                 "image": "netdata/netdata",
@@ -36,7 +36,7 @@
                 ],
                 "ports": {
                     "19999": {
-                        "description": "port used to access the webUI port",
+                        "description": "Port used to access the webUI port. MUST be 19999.",
                         "host_default": 19999,
                         "label": "webUI port",
                         "ui": true
@@ -44,12 +44,12 @@
                 },
                 "environment": {
                     "PGID": {
-                        "description": "GID of the docker group. Run `grep docker /etc/group | cut -d ':' -f 3` to find it.",
+                        "description": "GID of the 'docker' group. See System - Identity - Groups in Rockstor's UI to find it.",
                         "label": "PGID",
                         "index": 1,
-                        "default":472
+                        "default": 472
                     }
-				}
+                }
             }
         },
         "description": "Netdata is a scalable, distributed, real-time, performance and health monitoring solution for Linux, FreeBSD and MacOS.<p>Out of the box, it collects 1k to 5k metrics per server per second. It is the corresponding of running top, vmstat, iostat, iotop, sar, systemd-cgtop and a dozen more console tools in parallel. netdata is very efficient in this: the daemon needs just 1% to 3% cpu of a single core.</p><p>Based on the official docker image: <a href='https://hub.docker.com/r/netdata/netdata' target='_blank'>https://hub.docker.com/r/netdata/netdata</a>, available on x86_64/amd64, and aarch64/arm64.</p>",
@@ -59,6 +59,6 @@
         "ui": {
             "slug": ""
         },
-	    "version": "1.0"
+        "version": "1.0"
     }
 }
--- a/netdata.json
+++ b/netdata.json
@@ -2,12 +2,16 @@
     "Netdata": {
         "containers": {
             "netdata": {
-                "image": "titpetric/netdata",
+                "image": "netdata/netdata",
                 "launch_order": 1,
                 "opts": [
                     [
                         "--cap-add=SYS_PTRACE",
                         ""
+                    ],
+                    [
+                        "--security-opt",
+                        "apparmor=unconfined"
                     ],
                     [
                         "--net=host",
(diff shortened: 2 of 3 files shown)


phillxnet:
@FroggyFlox Thanks for tending to this.

Re:
> * The only drawback is the loss of the configuration override volume that was bound in our current rock-on, but I believe it isn't a big problem anyway.

completely agree. And would definitely be favourite to move to official image. Especially given their expanding cloud capability which depends on the client being up-to-date.

phillxnet:
@FroggyFlox I had completely forgot to mention and link here to our newly adopted preference for multi arch images, specifically favouring AArch64 in addition to the obvious/ubiquitous x86_64 of course.

Linking, as previously intended, to our related issue by @mcbridematt re arm64 rock-on coverage:

"[arm64] Discourage or prevent installation of Rockons that are not available on the machines architecture" #2191

So once this pr is merged we can update that canonical list/issue re netdata Arm64 support.

FroggyFlox:
@phillxnet , good thinking... I noticed the official image would bring aarch64 support but forgot about the list established by @mcbridematt.

FroggyFlox:
Thanks for testing this PR, @Hooverdan96!

Both your suggestions make perfect sense, so I'll include them when I can.

Specifying the architecture was something that came up once the incompatibility of some rock-ons for arm64 was raised, so that's why it's been thought as "indicating the lack of compatibility for arm64" rather than "indicating the supported architectures". You do raise a good point, though as that isn't the most intuitive approach for the vast majority of users, so I'm in favor in explicitly mentioning that both architectures are supported; as you suggest.

You're right that we should have a proper way to indicate this sort of things in the not-too-distant future, anyway, so we can always clean this kind of redundant information when it will be time for it.

Thanks!

Hooverdan96 (approved):

Hooverdan96:
@FroggyFlox, you added the "change requested" label recently after you pushed the latest changes. Is there anything else that needs to be done? I test installed your latest version and it's running nicely on my box now, but since I don't have any ARM devices available for testing I can't say anything about the other architectures.

One thing I was thinking about - in the tooltip for the PGID you are providing a command line entry to find out what the PGID for the docker group is ... in the spirit of simplicity, for a less technically inclined user, would it be better to point them to the WebUI to find the PGID under the "System --> Identity --> Group" menu path and to look at the GID column value next to the docker group?

As much as I love the tooltips on the Rock-ons, the fact that they're ephemeral makes it hard to take any detailed instructions from them (or, e.g. copy a command line entry or host address). Well, at least for me, as I'm old and my photographic memory is fading fast :).

FroggyFlox:
> Is there anything else that needs to be done?

Yes, for the same reason as highlighted recently in your Watchtower work. As a different image will be used, I need to create a new rock-on rather than updating the current one.




> One thing I was thinking about - in the tooltip for the PGID you are providing a command line entry to find out what the PGID for the docker group is ... in the spirit of simplicity, for a less technically inclined user, would it be better to point them to the WebUI to find the PGID under the "System --> Identity --> Group" menu path and to look at the GID column value next to the docker group?

Very good point, I'll include it.


> As much as I love the tooltips on the Rock-ons, the fact that they're ephemeral makes it hard to take any detailed instructions from them (or, e.g. copy a command line entry or host address). Well, at least for me, as I'm old and my photographic memory is fading fast :).

Yes, a proper write-up in our documentation is the preferred way to show this kind of information. I'm thinking of implementing a proper link to write-ups for rock-ons that have one. This way, we can easily inform the users of any installation recommendations/instructions. Ideally, we should have a write-up for all rock-ons available in the catalog.

Hooverdan96:
Well, duh ... you are absolutely right on the need for a new Rock-on. Like I said before, my memory ... :)

FroggyFlox:
> As a different image will be used, I need to create a new rock-on rather than updating the current one.

Following up on this requirement, I will now close this PR, and create split the work:
- Deprecate old Netdata rock-on
- Add new Netdata (official) rock-on

PR #239, opened 2020-10-28
Issue#233 Watchtower RockOn Update

### LINKED ISSUE #233: watchtower update needed
The watchtower rockon should point to https://hub.docker.com/r/containrrr/watchtower as the link has been updated for github and dockerhub.  It is the same image but we're pointing to a deprecated link.

Hooverdan96:
Fixes #233  Update to Watchtower RockOn.

N/A - existing RockOn updated
- name: Watchtower
- website: https://containrrr.dev/watchtower/
- description: automatic monitoring and updates of existing Rockon Docker images, available for ARM and AMD64

- docker image: https://hub.docker.com/r/containrrr/watchtower

### Checklist
- [X ] `"website"` object links to project's main website


Hooverdan96:
https://github.com/rockstor/rockon-registry/issues/233

FroggyFlox:
Thanks for stepping up to this one, @Hooverdan96 !

I actually had a look at that issue a little while ago and was wondering on how to best implement Watchtower as a rock-on, especially given the risks that using it has. I'm indeed worried about the risk of having a rock-on updated to a version that is less stable without the user wanting it. I actually noticed that Watchtower seems to include a workaround by offering a way to update only specific containers as instructed by the user:
https://containrrr.dev/watchtower/container-selection/

This uses labels, which Rockstor now supports, and even seems to allow for a container to be only "monitored" and not "updated" by Watchtower. What do you think, have you used this feature? I have not but it seems like a very good fit to me, and would not complicate the rock-on much. We can simply explain the labels to add, as well as pointing to the doc linked above in the "more_info" section. A "how-to" in the Rockstor documentation could also be helpful. Let me know what you think of the idea.

On another note, would you have some information on why the `--priviledged` flag is needed? As this can bring some security concerns, I'd rather make sure it is essential.

Thanks again for stepping up to this one!

Hooverdan96:
@FroggyFlox - I am not sure that the `--privileged` is required. It didn't seem to matter on my system.

I had rolled my own a while ago and exposed quite a few of the options that are listed in the documentation.
So, I have used the label to have some containers updated and some others excluded. We could add this in the environment section like this (excerpt from my version):
```

"WATCHTOWER_LABEL_ENABLE": {
	"description": "Update containers that have a com.centurylinklabs.watchtower.enable label set to true (you can use the 'Add label' functionality to add labels after installing Rockons). If this flag is set to false (default), containers with the com.centurylinklabs.watchtower.enable label set to false will be excluded.",
						"label": "Filter by label enable",
						"default": "false",
						"index": 8
```
Ignore the "default" of course ... this rockon was one of the reasons why I had asked about default values ... I only kept them in there, so I could remember what the default values were without going back to the documentation :).

So, yes, in my opinion it would make sense to at least add the Label enabler.

You can also adjust the polling interval or a CRON scheduling statement (but not both parameters, since the current RockOn structure requires them to be populated vs. optional). I used that because I didn't want it to go look for a new image too often (e.g. once a week or once a month only), to minimize disruptions:

```
					"WATCHTOWER_POLL_INTERVAL": {
						"description": "This value controls how frequently Watchtower will check for new images (in seconds). Default is 300 seconds.",
						"label": "Polling Interval (sec)",
						"default": "300s",
						"index": 5
					},
					"WATCHTOWER_SCHEDULE": {
						"description": "CRON expression (6 fields) defining when and how often to check for new images. Note: Either Polling Interval or scheduling should be defined, but not both",
						"label": "Scheduling (CRON format)",
						"index": 6
```
Finally, to keep the number of "unused" images small, here is a cleanup parameter, that I found useful

```
					"WATCHTOWER_CLEANUP": {
						"description": "Removes old Images after updating",
						"label": "Remove old images after update",
						"default": "false",
						"index": 1
```
And there are more that are conveniently exposed using the environment variables, but, some are mutually exclusive, so per our current design we have to either pick one, or none ...

FroggyFlox:
Thanks a lot for all the additional input, this is excellent!

I agree with you about the `"WATCHTOWER_LABEL_ENABLE"` ENV; I'm actually even wondering whether it wouldn't be preferable to "force" this as an "option" but that might be too much. What you have written here sounds like a good way to present it. I might personally put the "true" value as "recommended for stability reasons", while explaining that will require the user additional customization of all the other rock-ons that users will want monitored. Do you think that would be too much?

I also like your idea of customizing the polling interval/set a cron schedule. I'm surprised to see the default at 300 sec, that seems like a very short interval indeed! Regarding using a `POLL_INTERVAL` or `SCHEDULE` ENV, I'm not sure... I think I would prefer the schedule option as one can thus easily set a "once a week" or "once a month" like you said, rather than setting a very high value if one wants to use a monthly schedule. The only drawback, is that the CRON format is not as user-friendly as the polling interval. Maybe we could put an example for a weekly check and a link to some documentation on how to set that up?

I like the `CLEANUP` ENV too, that's definitely a useful option to surface.

Thanks again!

Hooverdan96:
The CRON gives more flexibility I agree. Since I don't use Cron very often I use something like this: https://www.freeformatter.com/cron-expression-generator-quartz.html#crongenerator. If you know of one that's you think is better, we can use that.
which I could reference in the description with the notice that the Year field does not apply.
I think we can leave the Label enable flag as an ENV and recommend that it is set to true and that it will require additional work on existing containers.
I believe, Watchtowers fundamental recommendation was to "run on demand" to make container updates easier, e.g. if you don't remember what the original settings were (which, with Rockstor is not an issue since they are transparent). But with the ability to schedule large(r) intervals it should alleviate the risk somewhat of just having container willy-nilly being updated.

If the above is ok, I will make the changes and push them again.

FroggyFlox:
> Since I don't use Cron very often I use something like this: https://www.freeformatter.com/cron-expression-generator-quartz.html#crongenerator.

Perfect! I was actually wondering about linking to something like that so that's great!

Thanks a lot for taking care of these changes. Have you ever noticed some incompatibility with Rockstor not picking up a rebooted/updated container or something like that, by any chance?

Hooverdan96:
@FroggyFlox alright, I will work on the changes.
I have not seen the incompatibility with Rockstor (didn't even think about that, to be honest). I will test this for a bit, after making the updates to see whether I notice an issue with that or not.

Hooverdan96:
@FroggyFlox - I made the changes (haven't pushed them yet), but it appears that with the --privileged flag the images are cleaned up afterwards (if the flag is set) and without they are not ... so not sure whether there is another option or not. Unless there is, I would tend to leave the privileged flag in the profile.

Hooverdan96:
@FroggyFlox - I pushed the latest version (somehow I had to force push, not sure why ... Git sometimes confuses me) with the latest changes we discussed.

FroggyFlox:
Thanks @Hooverdan96 , I'll try to have a look at it soon.

It seems we do need the `--privileged` flag for now, then, I agree. I would prefer not to have it in the future, though, so I think we can revisit that once we have a proper implementation of cleanup routines like that. We've talked a lot in the past about surfacing docker cleanup routines in Rockstor's webUI, so once we have that up and running, I think we should be in a good position to remove this `--privileged` flag here in an effort to lock down our containers as best as we can. The cleanup of unused docker images/containers could indeed be set by the user as a scheduled task, for instance...

On a side note, I noticed this PR includes your Logitech Media Server rock-on as well so that might be related to the issue you were having in pushing to this branch. Maybe that won't cause problems once we've taken care of that one and merged it, though, so we'll see. I just wanted to point that out to you for the moment.

Thanks again!

Hooverdan96:
@FroggyFlox - on the container cleanup, I think the ability to cleanup containers would be a good idea to have in the "general" docker/RockOn management as a task or ad-hoc (especially for users that roll their own RockOns frequently.
Then we could also remove the cleanup option in the Watchtower rockon. I suspect, any RockOn that "helps" manage docker containers will need privileged access ... but we will find out.
I am very sorry about the inclusion of the Logitech Rockon, I did not add it to this branch and when I checked the `git status` it didn't show as staged. Still, it somehow ended up in here ... you can tell I am not a frequent multi-PR contributor on github :)
Not sure, whether then the easiest will be to merge both the Pull Requests together once they are both reviewed?

FroggyFlox:
Thanks for your feedback, it's good to know it's a sensible idea; let's try to get to that once we can, then.

>  Not sure, whether then the easiest will be to merge both the Pull Requests together once they are both reviewed?
Not sure.... personally I would prefer to keep them separated as it helps keep track of any change, and we can easily revert if the needs arises. Now, I'm thinking of working on merging the LMS one first and then we can see exactly how to "fix" the Watchtower PR if any fix is indeed needed.

FroggyFlox:
@Hooverdan96 ,

I think this rock-on is ready but we need to fix the issue with the Logitech Media Server rock-on that is somehow included. You may want to close this PR, rebase your Hooverdan96:master branch on rockstor:master, and then re-commit the changes to `watchtower.json`. It might be cleaner this way than trying to exclude the commits that shouldn't be included.

It's all up to you in the end, however, so no worries and we'll make it work either way.

Thanks again for your continued contributions!

Hooverdan96:
@FroggyFlox, I will try to make it work with the rebasing and see that I don't screw it up more ... :)

Hooverdan96:
Closing this pull request for clean-up of commit. Will create a new one.

Hooverdan96:
@FroggyFlox - please check it out. I created a new Pull Request.

PR #246, opened 2020-11-13
Fixes #233 Watchtower Rock-on update (replaces Pull Request #239)

### LINKED ISSUE #233: watchtower update needed

Hooverdan96:
Fixes #233 Update to Watchtower RockOn.
Replaces Pull Request #239 

General information on project
N/A - existing RockOn updated

name: Watchtower
website: https://containrrr.dev/watchtower/
description: automatic monitoring and updates of existing Rockon Docker images, available for ARM and AMD64
Information on docker image
docker image: https://hub.docker.com/r/containrrr/watchtower
is an official docker image available for this project?: yes

Checklist
[X] Passes JSONlint validation
[ ] Entry added to root.json in alphabetical order (for new rock-on only)
[X] "description" object lists and links to the docker image used
[X] "description" object provides information on the image's particularities (advantage over another existing rock-on for the same project, for instance)
[X] "website" object links to project's main website


watchtower.json
```diff
@@ -1,23 +1,42 @@
...
+                    ]
+				],
+				"environment": {
+					"WATCHTOWER_LABEL_ENABLE": {
+						"description": "If this flag is set to 'true', containers with a specific label string (see the 'More info' icon) will be updated (you can use the 'Add label' functionality to add labels after installing Rockons). If the flag is set to 'false', containers with the label will be excluded. It is recommended to set it to true during installation. Note: Only running containers will be considered.",
+						"label": "Filter by label enable",
+						"index": 1
+					},
+					"WATCHTOWER_SCHEDULE": {
+						"description": "CRON expression (6 fields) defining when and how often to check for new images.",
```
FroggyFlox, line 29: We probably want to give an example here, for instance:
```suggestion
						"description": "CRON expression (6 fields) defining when and how often to check for new images. To run Watchtower every month on the 1st, at noon: 0 0 12 1 * ?",

```

Of course, change it to any other example if you think a different "default" schedule is preferred.

FroggyFlox (changes requested):
Only a minor change to this as everything looks good! I do have a major comment but I'll detail that in a separate post.

FroggyFlox:
Thanks again for this PR, @Hooverdan96 !

As you will see, I have only a small comment related to giving an example for a CRON config in the tooltip window.

On a different note, after testing the rock-on, I noticed a more important problem with our approach: I think we should make a new rock-on (Watchtower official) rather than updating the current one. Indeed, in the current state, while updating the list of available rock-ons works without a problem for users who do not have the "old" already installed, it causes issues for those who do have it installed. This is related to how Rockstor deals with that and might use some improvement in the future, but in the meantime, this represents a problem for users in this situation. While they can indeed uninstall the rock-on and then re-update the list of available rock-ons, I think it's better to not force something like that and avoid creating an error.

I apologize for not thinking about that issue earlier; I learned something new today.

So in the end, I think we should create a new rock-on named "Watchtower (official)" (or something like that) instead of updating the current one. Let me know what you think of it.

watchtower.json
Hooverdan96: yep, good idea, so the user can more quickly set this up and later think about other schedules (that's where a change without uninstall/reinstall would be handy).

Hooverdan96:
Ah, yes, it has to do with how information is retained in the database, correct? And, yes, it would make sense to have some "upgrade" functionality there eventually that would make this types of updates less intrusive (I guess background process that uses the rock-on delete program, while retaining the config parameters and then automatically reinstalling or something to that effect).
I am mostly wondering whether we confuse users more, if we have multiple versions of the same Rock-on, to decide whether to pick this one or the other version of the Rock-on.
Would it then make sense to update the other Watchtower Rock-On with a title that's something like "Watchtower (Legacy)" to indicate that new installations are recommended to switch to the official container? Or is it better to deprecate the old one and introduce the new one instead?
So, if you feel strongly about making this a new one instead of replacing the existing one, how do we go about it? Close this pull request again, create a new one with an altered json (new name, new container label, etc. and then either deprecate the other one, or leave them both in place?

FroggyFlox:
That's my bad, I did mean to detail that it would still be a replacement but forgot to clearly state it. Yes, we would create a new "Watchtower (official)" rock-on and then deprecate the older one so that there is no confusion. No need to keep a rock-on if its image is no longer maintained; it's actually something we've been trying to clean up (and with which you helped immensely).

Ideally, I would thus try to close this PR, and make two new ones:
- create a new rock-on Watchtower (official)
- deprecate the current Watchtower rock-on

I apologize again for making this so complex; I actually learned something I didn't think of in the process so hopefully that'll make things smoother in the future.

Hooverdan96:
I'll become the guy with the most opened and closed pull requests for the same item on this repo :). Sounds good. I am closing this one and will open first one to create the new one, and then another one subsequently for the deprecation (I guess, I shall wait for the second one, until the first one is merges, as both will involve the root.json files.

PR #263, opened 2020-12-08
Seafile v7+

s-bernard:
This pull request proposes to add a new rock-on Seafile v7+ to get Seafile version 7 and above.
Tested on rockstor 4.0.4.

It also renames current Seafile rock-on to Seafile v6 (should I rename the container also?)

- docker image: seafileltd/seafile-mc
- official image for seafile 7+

### Checklist


seafile7+.json
```diff
@@ -0,0 +1,61 @@
...
+          ["-e", "DB_HOST=mariadb"],
+          ["-e", "DB_ROOT_PASSWD=MDA0NDRmNjYxOTMyMzJiOTY5MzZkNGY4"],
+          ["-e", "SEAFILE_SERVER_LETSENCRYPT=false"],
+          ["--link", "seafile-mariadb:mariadb"],
+          ["--link", "seafile-memcached:memcached"]
+        ]
+      }
+    },
+    "version": "latest",
+    "description": "Secure file sharing and hosting. <p>Based on the official <a href='https://hub.docker.com/r/seafileltd/seafile-mc' target='_blank'>Seafile multi-container image</a> for versions 7 and above. Use MariaDB and Memcached as separate container.</p> <p>HTTP only, to add HTTPS, you can use Nginx-proxy-manager.</p>",
```
FroggyFlox, line 53: Add supported architecture:
```suggestion
    "description": "Secure file sharing and hosting. <p>Based on the official <a href='https://hub.docker.com/r/seafileltd/seafile-mc' target='_blank'>Seafile multi-container image</a> for versions 7 and above, available for amd64/x86_64 architecture only. Use MariaDB and Memcached as separate container.</p> <p>HTTP only, to add HTTPS, you can use Nginx-proxy-manager.</p>",
```

FroggyFlox (changes requested):
@s-bernard ,
Thanks a lot for such an excellent PR and rock-on! It works flawlessly in my hands. I only have a minor suggestion related to explicitly stating the supported architecture; we now try to consistently specify this for all rock-ons while we implement a more automated way to do so.

FroggyFlox:
Thanks a lot for an excellent PR, @s-bernard !

As mentioned in my review, I only have one minor suggestion and I'll merge this once included (after one final test).

> It also renames current Seafile rock-on to Seafile v6 (should I rename the container also?)

I don't think we need to, and I like your current renaming. I'm not a Seafile user, so I'm actually curious about the differences between Seafile 6 and 7... My ultimate question is whether or not we need to keep the "old" Seafile 6 rock-on available. We indeed are now in an undergoing effort to reduce unnecessary multiple versions for a given project so I would appreciate your input (as a Seafile user) on whether or not Seafile v6 should be kept around or not. I personally think we can keep it for a little while but plan to deprecate it in the not too distant future. Of course, if both versions are so different that some users would still require v6, then we can keep it for a little longer.

Speaking of migrating between versions, based on the volume bindings between v6 and v7 it seems that the same share can be used for the `/shared` volume in both versions, meaning that the v7 rock-on can be a drop-in replacement for the v6 rock-on... Would I be correct in thinking that?

s-bernard:
Seafile 7 is just an simple evolution of Seafile 6. There is no need to keep version 6 forever.
I kept the previous rock-on because the migration between the two request a lot of manual operations: https://manual.seafile.com/docker/6.3%20upgrade%20to%207.0/ (This is due to the change of docker architecture, multiple containers for v7 vs everything in one for v6.)

There is the /shared volume which is the same for both version but v6’s /shared also contains mariadb data which have to be extracted and set in their own volume + configure mariadb to work remotely and not just in local mode (see link above).

FroggyFlox (changes requested):
@s-bernard, thanks a lot for making the changes, and sorry for the delay on my end.

I tested it again and everything still works like a charm for the Seafile v7+ rock-on. On my limited test hardware, it takes a while before the seafile-mc container is fully booted up and ready to serve (a little more than 4 minutes), but that was also the case for the older Seafile rock-on.

I did encounter an issue when testing the case for users who currently run the Seafile rock-on. Indeed, we run into an error while trying to fetch the metadata for the Seafile v6 rock-on due to the conflicting container name `seafile` already belonging to the old Seafile rock-on. I see the following options to resolve this:
**a**. update the container name in the old rock-on to `seafile6` in addition to the changes you already made to this json file.
**b**. Revert the changes you made to the old Seafile json file, but alter its `"description"` field to reflect the fact that this rock-on will be deprecated in the near future... something like:
```
        "description": "<strong>Important:</strong> This rock-on runs an out-of-date Seafile version (v6) and will be deprecated in the near future. We recommend switching to the Seafile v7+ rock-on instead. <p>Secure file sharing and hosting. Based on the official Seafile docker image: <a href='https://hub.docker.com/r/seafileltd/seafile' target='_blank'>https://hub.docker.com/r/seafileltd/seafile</a>.",
```

I personally would favor option **b** as we thus avoid altering any of the critical metadata for a rock-on that users may have currently running.

Let me know what you think of it, and thanks again for your continued effort!

s-bernard:
I agree with option b so it does not break for existing users. I’m also adding a link to migration documentation in the description.

s-bernard:
Will it breaks if I still do the file renaming `Seafile.json` => `seafile6.json` without changing the "title" Seafile?

s-bernard:
I applied option b but I kept the file renaming without changing the metadata. Tell me if it’s OK.

FroggyFlox:
Hi @s-bernard,

First, let me apologize for such a delay in finishing reviewing your PR... Lief got in the way and I've also been trying to wrap up a couple branches on which I've been working in the rockstor-core repo in order to get Rockstor even closer to the v4 release. I've finally been able to test the following:

> I'll test that again and verify this point anyway.

It does work well; for instance, here's what I have on a system with the old Seafile rock-on already installed. Updating the list of available rock-ons updated everything about it (and thus surfacing its "deprecation notice" and instructions on how to migrate), and I could install successfully the Seafile v7+ rock-on in parallel.

image


I noticed a conflict in the `root.json` file but that's related to another PR that was merged after your PR submission so I'll try to resolve that one.

Thanks again for such a nice rock-on!

FroggyFlox (approved):

PR #271, opened 2021-02-11
Fix nginx-proxy-manager for rockstor 4

Marenz:
As IPv6 is currently disabled, we also have to disable it for this rock-on via environment variables otherwise nginx will not start:

```
2021/02/11 20:35:31 [emerg] 14860#14860: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:32 [emerg] 14869#14869: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:33 [emerg] 14878#14878: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:34 [emerg] 14887#14887: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:35 [emerg] 14896#14896: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:36 [emerg] 14905#14905: socket() [::]:8080 failed (97: Address family not supported by protocol)
2021/02/11 20:35:37 [emerg] 14914#14914: socket() [::]:8080 failed (97: Address family not supported by protocol)
```
fyi @DarkYamik

### DIFF
--- a/NginxProxyManager.json
+++ b/NginxProxyManager.json
@@ -4,6 +4,12 @@
 			"nginx-proxy": {
 				"image": "jlesage/nginx-proxy-manager",
 				"launch_order": 1,
+				"opts": [
+					[
+						"-e",
+						"DISABLE_IPV6=true"
+					]
+				],
 				"ports": {
 					"8181": {
 						"description": "Admin WebUI port",


DarkYamik:
Thank you @Marenz for this! I'm still using Rockstor 3 so I did not notice this!

FroggyFlox:
Thanks a lot, @Marenz, and nice catch!

Am I correct in understanding that the `DISABLE_IPV6` environment variable **should always** be set to `false`, then?
If yes, I think we should move it to the `"options"` object so that we don't add another field during the install wizard that might make the user wonder what the value should be.

Hooverdan96:
@FroggyFlox, @Marenz if I understand the documentation correctly, `DISABLE_IPV6` should for all Rockstor versions always be set to `1` (i.e. "true" for disabling ipv6 support).
And since Rockstor is not using `ipv6` (don't know whether forever, or only foreseeable future) then it would make sense to me as well to move it to the `options` section in the json file.

Marenz:
Updated to be part of the `opts` array.

FroggyFlox (approved):

FroggyFlox:
Thanks a lot, @Marenz  for the additional fix; I just tested it again and it now boots up without issue. Thank you for tending to keeping this important rock-on in shape and working order!

@Hooverdan96, thanks a lot for your input on the resolution, always very helpful!

I'll go ahead and merge this PR, followed shortly by its publication.

PR #272, opened 2021-02-19
Added NZBHydra2, as linuxserver.io have deprecated older hydra image

thedrjones:
When looking at the pages on the Docker Hub for [linuxserver.io/hydra](https://hub.docker.com/r/linuxserver/hydra/), it mentions in the Application Setup section that that image is deprecated, pointing the user to [linuxserver/nzbhydra2](https://hub.docker.com/r/linuxserver/nzbhydra2) instead. I have been using a self-defined image for NZBHydra2 for some time, so I have tidied up the configuration and shared here.

Note: NZBHydra2 no longer allows you to import you existing NZBHydra configuration.

- name: linuxserver.io NZBHydra2
- website: https://github.com/theotherp/nzbhydra2
- description: NZBHydra2 is meta search application for NZB indexers, offering a unified search experience over several different indexers, it can also integrate into your usenet client. Like many linuxserver.io images, it supports x86_64, arm64v8 and arm32v7 

- docker image: https://hub.docker.com/r/linuxserver/nzbhydra2
- is an official docker image available for this project?: No official image, but linuxserver is in the recommended suppliers of images by the project author.

### Checklist

### DIFF
--- a/NZBHydra2.json
+++ b/NZBHydra2.json
@@ -1,0 +1,48 @@
+{
+    "NZBHydra2": {
+        "containers": {
+            "nzbhydra2": {
+                "image": "linuxserver/nzbhydra2",
+                "launch_order": 1,
+                "ports": {
+                    "5075": {
+                        "description": "NZBHydra2 WebUI port. Suggested default: 5075",
+                        "host_default": 5075,
+                        "label": "WebUI port",
+                        "protocol": "tcp",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/config": {
+                        "description": "Choose a Share for NZBHydra2 configuration. Eg: create a Share called nzbhydra-config for this purpose alone.",
+                        "label": "Config Storage"
+                    },
+                    "/downloads": {
+                        "description": "Choose a Share for NZBHydra downloads. Eg: create a Share called nzbhydra-downloads for this purpose alone.",
+                        "label": "Download Storage"
+                    }
+                },
+                "environment": {
+                    "PUID": {
+                        "description": "Enter a valid UID to run nzbhydra as. It must have full permissions to all Shares mapped in the previous step.",
+                        "label": "UID to run nzbhydra as.",
+                        "index": 1
+                    },
+                    "PGID": {
+                        "description": "Enter a valid GID to use along with the above UID. It(or the above UID) must have full permissions to all Shares mapped in the previous step.",
+                        "label": "GID to run nzbhydra as.",
+                        "index": 2
+                    }
+                }
+            }
+        },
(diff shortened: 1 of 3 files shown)


FroggyFlox (changes requested):
I noticed the port used for the webUI needed updating (needs to be 5076 and not 5075) so I made some suggestions accordingly.
Besides that, I've only found a couple of style suggestions and all seems to work fine on both x86 and arm64 machines!

FroggyFlox:
Hi @thedrjones, and thanks for your submission!

As you can see in my review, I just noticed that the webUI port needs to be updated otherwise we can't access the webUI. Otherwise, it all seems to be working as intended.

On another note, I think we should add a deprecation notice to our current NZBHydra rock-on, as it would help orient our users before we actually remove that rock-on. See a recent similar case for our Seafile rock-on: https://github.com/rockstor/rockon-registry/commit/0e27e30aacde6f9719feb046e54f1a8cc007b204
You can write something similar to:
> the docker image used by this rock-on has been deprecated by its authors and will thus be removed from Rockstor in the near future. We recommend you to use the NZBHydra2 rock-on instead.

Thanks again for your contributions!

thedrjones:
Hi @FroggyFlox 
I never noticed that 5076 is the default port on that image ... I guess my configuration directory has 5075 baked in.
Suggestions committed.

thedrjones:
I've also added a deprecation message to the older image too, and fixed a mix of tabs and spaces. Let me know if you want it separating again.

NZBHydra2.json
```diff
@@ -0,0 +1,47 @@
...
+                        "index": 2
+                    }
+                }
+            }
+        },
+        "description": "NZBHydra2 is a meta search for NZB indexers. <p>Uses custom docker image by <a href='https://www.linuxserver.io/' target='_blank'>Linuxserver.io</a>: <a href='https://hub.docker.com/r/linuxserver/nzbhydra2/' target='_blank'>https://hub.docker.com/r/linuxserver/nzbhydra2/</a>, available for amd64, arm32v7 and arm64v8.</p>",
+        "ui": {
+            "slug": ""
+        },
+        "volume_add_support": false,
```
FroggyFlox, line 43: @thedrjones , is there a good reason to prevent the addition of volumes here? I'm not a user of this project so I don't know how it works, but do you think some users might want to bind their existing media library, for instance? If yes, we might want to allow that. Let us know what you think.

FroggyFlox (commented):
@thedrjones , I still need to spend time to properly test it but I just one question (see comment) in the meantime.

Thanks again for tending to updating this rock-on!

NZBHydra2.json
thedrjones: I have never needed to bind another set of shares to it over the years, so that's why I never enabled it :).
NZBHydra is just a search aggregator that allows you to combine searches on your configured newsgroup indexers (e.g sites like NewsBin, NZBGeek) through one unified web page (rather than checking out several indexers by hand for the same search, it can then weight those results to things you may consider more reliable higher up the list).
It can also allow you to send the result nzb metadata file over to your newsgroup client (e.g. SABnzbd) for that client to do the download. NZBHydra itself does not actually need to download into or touch your media libraries.

FroggyFlox: Makes sense, thanks a lot for the information!

FroggyFlox (approved):
Thanks @thedrjones,

It all works as intended on both x86_64 and Arm64 architectures so I'm going to go ahead and merge it now. Thanks again!

PR #279, opened 2021-03-17
SiliconDust's HdHomerun standalone DVR engine Rockon 

### LINKED ISSUE #278: Add HDHomeRun DVR engine Rockon
HDHomeRun Tuners can record (with a valid subscription) live tv using a separate DVR engine that can be installed under windows or Linux. This makes a good Rockstor Rockon to add broadcast TV recordings to one's media library.

Hooverdan96:
Fixes #278  .

- name: HDHomeRun DVR engine from SiliconDust
- website: https://www.silicondust.com
- description: standalone DVR engine that work in conjunction with SiliconDust's HDHomeRun series of tuners that don't have the DVR built-in. Requires a valid subscription and recent HDHomeRun hardware

- docker image: https://hub.docker.com/repository/docker/jackdock96/hdh_dvr
- is an official docker image available for this project?: Not at this time. I created this docker image to self-update with the latest engine and also offer multi-arch images to support Rockstor's direction of installing on ARM architecture systems, so the docker container doesn't have to be maintained in the repository too frequently.
- **Note**: because the dvr engine uses multi-cast the Rockon is currently configured to use the `net=host` option. Unless one implements a `macvlan` or uses tools like pipework, multi-cast from LAN to container and vice-versa is not possible. I assume, Rockstor's Rocknet might be able to achieve that, but until we are completely cutover to 4.x on OpenSuse, this is the only option.
- Plan is to switch to an official image if/when it becomes available

### Checklist

### DIFF
--- a/hdhomerun_dvr.json
+++ b/hdhomerun_dvr.json
@@ -1,0 +1,47 @@
+{
+  "HDHomeRun DVR Engine": {
+    "containers": {
+      "hdhomerun_dvr_engine": {
+        "image": "jackdock96/hdh_dvr",
+        "opts": [
+          [
+            "--net=host",
+            ""
+          ]
+        ],
+        "launch_order": 1,
+        "volumes": {
+          "/dvrdata": {
+            "description": "Choose a Share for HomeRun DVR engine configuration. E.g.: create a Share called hdh-config for this purpose alone. It is important that the owner/group specified later have full access to the share (R/W/E). Recommended size is a few MB if quotas are used",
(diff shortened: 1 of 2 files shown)


Hooverdan96:
The amd64 version works very well on my environments. I have not been able to test the ARM architecture container versions on "real" hardware at this time, though.

Hooverdan96:
Looks like all my rebasing ended up in this pull request ... that wasn't the intention. I closed this and will attempt to clean that up...

PR #283, opened 2021-03-19
Logitech Media Server (Squeezebox), move to latest release tag

### LINKED ISSUE #282: Logitech Media Server (Squeezebox) - upgrade to latest stable
The underlying docker structure has been adjusted to offer the latest stable build under the "latest" tag (has now entered the 8.x version). Replacing the current version 7.9.4 with a new one would be advisable. Based on past discussions, it should be introduced as a new one (especially, since the current specifies a version in its title, and the old one should be deprecated.

Adding additional information on how to continue to update/upgrade to the latest stable version, as the underlying container doesn't do it upon a stop/restart, like many of the ones that e.g. linuxserver.io offers.

Hooverdan96:
Fixes #282.

- name: Logitech Media Server
- website: https://mysqueezebox.com
- description: Replace current 7.9.4 LMS Rockon with one that uses "latest" label, so upon installation the latest stable version of LMS is pulled, instead of being pinned to an aging version number. Once this is accepted/merged the older version would be deprecated.

- docker image: https://hub.docker.com/r/doliana/logitech-media-server
- is an official docker image available for this project?: No, however this is a popular docker container (10+M pulls) and also used in the already existing Rockon. This newer version will also add some more explanation for updating in the More Info tag.
- this still requires the `net=host` option, because some of the plug-ins require multi-cast (e.g. the apple bridge, as well as some of the other plugins). Once docker is capable of routing multi-cast correctly, this option can be dropped. Alternatively, once Rockstor is completely migrated to OpenSuse, the Rocknet might be able to support this. If so, then we can change/update this one (as well as others without the host network option), but it will make installation also more involved ...

### Checklist

### DIFF
--- a/lms_latest.json
+++ b/lms_latest.json
@@ -1,0 +1,54 @@
+{
+	"Logitech Media Server - Latest Stable": {
+		"description": "Logitech Media Server (aka Squeezebox, slimserver, etc.). Latest stable version. For future updates see more information icon after first installation<p>Uses custom docker image: <a href='https://hub.docker.com/r/doliana/logitech-media-server' target='_blank'>https://hub.docker.com/r/doliana/logitech-media-server</a>. Available for amd64, arm/v7 &arm64 architecture.",
+		"more_info": "<h4>Update to newer version</h4><p>After first installation: to update to a newer stable version, either uninstall/reinstall the Rockon or use the Watchtower Rockon to manage the update of the underlying Media Server image, as it does not automatically update itself during a stop/restart of the Rockon like some other Rockons do. Don't use the update option within the LMS!</p><p>The LMS configuration files will remain in place on the volumes that were specified during the initial installation</p>",
+		"website": "https://mysqueezebox.com/",
+		"version": "2.2",
+		"ui": {
+			"slug": ""
+		},
+		"containers": {
+			"lms8": {
+				"image": "doliana/logitech-media-server",
+				"tag": "latest",
+				"launch_order": 1,
+				"opts": [
+					[
+						"--net",
+						"host"
+					]
+				],
+				"ports": {
+					"9000": {
+						"description": "WebUI port. Suggested default: 9000",
+						"host_default": 9000,
+						"label": "WebUI port",
+						"protocol": "tcp",
+						"ui": true
+					},
+					"9090": {
+						"description": "Command Line Interface (used for example by Android app Squeezer). Suggested Default: 9090",
+						"host_default": 9090,
+						"label": "CLI port",
+						"protocol": "tcp"
+					},
+					"3483": {
+						"description": "control channel for LMS (display, IR, etc.), tcp and udp (SB --> Slimserver discovery and for older SLIMP3 players). Suggested Default: 3483",
+						"host_default": 3483,
+						"label": "Control Channel"
+					}
+				},
+				"volumes": {
+					"/srv/squeezebox": {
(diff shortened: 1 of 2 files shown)


lms_latest.json
```diff
@@ -0,0 +1,54 @@
+{
+	"Logitech Media Server - Latest Stable": {
+		"description": "Logitech Media Server (aka Squeezebox, slimserver, etc.). Latest stable version. For future updates see more information icon after first installation<p>Uses custom docker image: <a href='https://hub.docker.com/r/doliana/logitech-media-server' target='_blank'>https://hub.docker.com/r/doliana/logitech-media-server</a>. Available for amd64, arm/v7 &arm64 architecture.",
+		"more_info": "<h4>Update to newer version</h4><p>After first installation: to update to a newer stable version, either uninstall/reinstall the Rockon or use the Watchtower Rockon to manage the update of the underlying Media Server image, as it does not automatically update itself during a stop/restart of the Rockon like some other Rockons do. Don't use the update option within the LMS!</p><p>The LMS configuration files will remain in place on the volumes that were specified during the initial installation</p>",
```
FroggyFlox, line 4: Similar to the previous LMS Rock-on you worked on, maybe we should mention that the webUI port MUST be set to 9000. You had the following mention in the previous LMS Rock-on:
```
<p>Due to some current <a href='https://github.com/rockstor/rockstor-core/issues/2209' target='_blank'>limitations in rock-on networking</a>, the web UI port of this rock-on must be set as 9000 and will thus prevent the installation of other rock-ons needing this port. Alternatively, the web UI can be accessed manually at http://rockstor-ip:9000</p>
```

lms_latest.json
```diff
@@ -0,0 +1,54 @@
...
+				"launch_order": 1,
+				"opts": [
+					[
+						"--net",
+						"host"
+					]
+				],
+				"ports": {
+					"9000": {
+						"description": "WebUI port. Suggested default: 9000",
```
FroggyFlox, line 23: As with the previous LMS Rock-on, we should include that it MUST be set to 9000.
```suggestion
						"description": "WebUI port. This MUST be set to 9000",
```

FroggyFlox (changes requested):
Hi @Hooverdan96,

Thanks a lot for working to keep this Rock-on up-to-date. I completely agree with you on using the `latest` tag and creating a new Rock-on for it and then deprecating the previous one.
I'll try to find some time to properly test it on both architectures supported but in the meantime, I think we should include similar mentions as what we added in the previous Rock-on with regards to the fact that the webUI port must be set to 9000 as it uses host networking.

Once we have that in and once I can verify it works as intended on both architectures, I'll merge this in.

Thanks again for yet another crucial update from you; sorry I couldn't get to spend time on it sooner.

FroggyFlox:
Thanks a lot! I'll try to give it a final test asap on multiple architecture and Leap versions.

I don't think we need to add justification in the port description field as I feel it would be too much information in that particular spot, which would thus be counterproductive.

Thanks a lot for such a nice contribution and well-written PR; always a pleasure to work on yours! Sorry again it took me so long to find some time to look at it.

FroggyFlox (approved):
@Hooverdan96 , I've now tested it and can confirm it installs, runs, can be turned OFF, uninstalled, and re-installed on:
- Rockstor 4.0.9 on Leap15.3 x86_64
- Rockstor 4.0.9 on Leap15.2 Raspberry Pi4
- Rockstor 4.0.9 on Leap15.3 Raspberry Pi4

Thanks again for yet another excellent contribution!

PR #287, opened 2021-06-19
Create Gotify.json

freaktechnik:
Fixes #- .

- name: Gotify
- website: https://gotify.net
- description: Personal push notification service, similar to Pushbullet. Can also be used for https://unifiedpush.org

- docker image: https://hub.docker.com/r/gotify/server

### Checklist

### DIFF
--- a/Gotify.json
+++ b/Gotify.json
@@ -1,0 +1,39 @@
+{
+    "Gotify": {
+        "description": "a simple server for sending and receiving messages",
+        "ui": {
+            "slug": ""
+        },
+        "website": "https://gotify.net/",
+        "version": "1.0",
+        "containers": {
+            "gotify": {
+                "image": "gotify/server",
+                "launch_order": 1,
+                "ports": {
+                    "80": {
+                        "description": "Gotify server port. Suggested default: 80",
+                        "host_default": 80,
+                        "label": "Server port",
+                        "protocol": "tcp",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/var/gotify/data": {
+                        "description": "Choose a Share for the database, plugins and images for applications.  Eg: create a Share called gotify for this purpose.",
+                        "label": "Database and App Images",
+                        "min_size": 1073741824
+                    }
+                },
+                "environment": {
+                    "TZ": {
+                        "label": "Timezone",
+                        "description": "Example: Europe/Berlin"
+                    }
+                }
+            }
+        },
+        "more_info": "<p>Default username is <code>admin</code> and password is <code>admin</code>. For usage instructions visit <a href=\"https://gotify.net/docs/first-login\">the gotify Getting Started guide</a></p>"
+    }
+}
--- a/Gotify.json
+++ b/Gotify.json
@@ -1,11 +1,11 @@
 {
     "Gotify": {
-        "description": "a simple server for sending and receiving messages",
+        "description": "A simple server for sending and receiving messages. <p>Based on the official Gotify x86 docker image: <a href='https://hub.docker.com/r/gotify/server' target='_blank'>https://hub.docker.com/r/gotify/server</a>.</p>",
         "ui": {
             "slug": ""
         },
         "website": "https://gotify.net/",
-        "version": "1.0",
+        "version": "latest",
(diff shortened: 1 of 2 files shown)


FroggyFlox:
Sounds good, @freaktechnik; I'll try to find some time asap to give it a proper test/review.

Gotify.json
```diff
@@ -0,0 +1,39 @@
+{
+    "Gotify": {
+        "description": "A simple server for sending and receiving messages. <p>Based on the official Gotify x86 docker image: <a href='https://hub.docker.com/r/gotify/server' target='_blank'>https://hub.docker.com/r/gotify/server</a>.</p>",
```
FroggyFlox, line 3: Harmonize description with current guidelines:
```suggestion
        "description": "A simple server for sending and receiving messages. <p>Based on the official Gotify Docker image: <a href='https://hub.docker.com/r/gotify/server' target='_blank'>https://hub.docker.com/r/gotify/server</a> available for x86_64 architecture only.</p>",
```

Gotify.json
```diff
@@ -0,0 +1,39 @@
...
+        },
+        "website": "https://gotify.net/",
+        "version": "latest",
+        "containers": {
+            "gotify": {
+                "image": "gotify/server",
+                "launch_order": 1,
+                "ports": {
+                    "80": {
+                        "description": "Gotify server port. Suggested default: 80",
```
FroggyFlox, line 15:
```suggestion
                        "description": "Gotify server port. Suggested default: 8100",
```

Gotify.json
```diff
@@ -0,0 +1,39 @@
...
+                        "host_default": 80,
```
FroggyFlox, line 16: While setting this to `80` does work as Rockstor tries the next available port, it can be risky to leave it set at 80. In my case, for instance, Rockstor found the next one available to be 87 and while it does not break the Rock-On install, it does prevent access to Gotify's UI. Indeed, Firefox blocked access to it as port 87 is reserved for another protocol.
I'll file an issue to improve our Rockstor deals with this, but to circumvent this for the moment, I would recommend setting a `host_default` value to a high port number such as 8100. You can set it to whatever else you want in that "high number" space, though.
```suggestion
                        "host_default": 8100,
```

Gotify.json
```diff
@@ -0,0 +1,39 @@
...
+                    }
+                },
+                "volumes": {
+                    "/app/data": {
+                        "description": "Choose a Share for the database, plugins and images for applications.  Eg: create a Share called gotify for this purpose.",
+                        "label": "Database and App Images",
+                        "min_size": 1073741824
+                    }
+                },
+                "environment": {
```
FroggyFlox, line 29: No need for setting the TZ variable here, I believe, as Rockstor automatically binds the `/etc/localtime` volume in all containers:
https://github.com/rockstor/rockstor-core/blob/3ad7eee27a40badab7edeea6dc198e800f6b121e/src/rockstor/storageadmin/views/rockon_helpers.py#L187-L188

freaktechnik: I'm not familiar enough with Alpine to know if that actually works there. From quick googling it seems that's not always enough for Alpine?

FroggyFlox: Oh good point... I'll have to try that, then. Thanks for pointing that out.

freaktechnik: Thank you for looking into it. Somehow I had come to the conclusion that the image was alpine based.

wwwizzarrdry:
Can't pull the latest image due to a missing signature key.
`CommandException: Error running a command. cmd = /usr/bin/docker pull gotify/server:latest. rc = 1. stdout = ['']. stderr = ['Error response from daemon: missing signature key', '']`

_Version: RockStor 3.9.2-57 Linux: 4.12.4-1.el7.elrepo.x86_6_

Hooverdan96:
@wwwizzarrdry as mentioned by @FroggyFlox in reponse to your Rockstor community forum post:

https://forum.rockstor.com/t/gotify-missing-signature-key-failed-to-install-in-the-previous-attempt/10485

this is very likely related to the outdated version of docker, which in the underlying, now legacy CentOS cannot easily be further updated.

PR #292, opened 2021-08-25
Add a virtual machine host rockon using libvirtd

kageurufu:
This rockon runs libvirtd in docker, using KVM. 

This requires host processor support, you can check for support with `grep -E '(vmx|svm)' /proc/cpuinfo`. 
You also need nested virtualization support, Rockstor seemed to already have this enabled for `kvm_intel`, but it could be guaranteed with the kernel cmdline `kvm-intel.nested=1 kvm-amd.nested=1`. 

Completely untested, but ARM hosts could work as well. I don't have any relevant hardware to test on though.

Using virt-manager can be a little "crunchy" sometimes, pasting into VMs is nigh-impossible right now.

I'd like some other people to test this, but it works for me.

I'm uncertain how some of the checklist here should be handled, I figure I would open this to discussion

- name: virt-manager
- website: https://libvirt.org
- description: A virtual machine host and web interface

- docker image: https://hub.docker.com/r/hrishikesh/libvirtd, https://hub.docker.com/r/mber5/virt-manager
- is an official docker image available for this project?: not really

### Checklist

### DIFF
--- a/root.json
+++ b/root.json
@@ -78,6 +78,7 @@
     "Ubiquiti Unifi linuxserver.io": "unifi-linuxserver.json",
     "Unifi Controller": "unifi.json",
     "uTorrent": "uTorrent.json",
+    "virt-manager": "virt-manager.json",
     "Watchtower offical": "watchtower_official.json",
     "Xeoma Video Surveillance": "xeoma.json",
     "YouTrack official": "youtrack-official.json",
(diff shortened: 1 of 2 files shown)


phillxnet:
@kageurufu From the forum discussion it looks like you were abandoning this approach for now. Consequently I'll close this pr by way of tidying house. If you end up re-visiting then just pop in a new pr based on current repo content and we can pick up where we left off. It would also be good to link to this pr in that case; if that is how things pan out.

Thanks again for sharing your work thus far, quite the interesting project but likely outside the remit / scope of our Rock-ons actually.

Issue #185, opened 2019-04-13
remove existing discourse rock-on


phillxnet:
The existing image used discourse/discourse is now superseded, and presumably deprecated by: discourse/base. However given the additional 'special treatment' this newer docker image requires a companion issue has been opened in rockstor-core. This companion issue, linked below, is a dependency of this issue and so would need to be completed prior to this issue receiving attention.

Further input on this and the below linked rockstor-core issue is welcome, as always, as this issues author only has passing familiarity with discourse setup.

Associated dependency in rockstor-core:
https://github.com/rockstor/rockstor-core/issues/2033

mcbridematt:
As noted in #228, it is not possible to pull "discourse/discourse" anymore as the default ":latest" container has been removed from Docker Hub.

phillxnet:
@mcbridematt Thanks for the update here. I'm moving more towards just removing this rock-on entirely actually as given it's broken now and the work to re-intergrate it is non trivial I move to us removing it. Quick and better all around. Me and @FroggyFlox have been intending to do a bit of Rock-on house keeping in like with just removing all obviously broken Rock-ons anyway. Anyone interested enough can always step up and re-add it along with it's required special treatment if they fancy. But for now I vote for removal.

@FroggyFlox & @mcbridematt do you concur?

We could leave the rockstor-core issue open for those wanting to have a go at this as it indicates the additional special treatment element. And close this issue on the existing broken Rock-on's removal.

phillxnet:
@FroggyFlox I've just noticed this issue again.

Lets just remove this rock-on. It's currently broken anyway and given the info in this issue, and the lack of feedback we've gotten on it's failure it won't be missed. Plus the non trivial effort to support it's 'special' nature is not currently within our scope. We can, as stated earlier, leave the rockstor-core element as a guide for anyone wanting to re-add; although it will of course live on in git history so we could similarly remove that also.

phillxnet:
I am currently working on this issue.

phillxnet:
Closing as:
Fixed by #299

PR #299, opened 2021-11-23
remove existing discourse rock-on #185

phillxnet:
The existing image used discourse/discourse is now superseded, and presumably deprecated by: discourse/base, which needs quite a bit of 'special' treatment. And given we have had almost no reports of our existing failure due to a non working image lets just remove the Rock-on entirely.

Fixes # 185

### Checklist
- [x] Entry removed to `root.json` in _alphabetical_ order (for new rock-on only)


phillxnet:
@FroggyFlox Given this is a straight, long overdue removal. I'll go ahead and merge directly.

PR #280, opened 2021-03-17
Add SiliconDust's HdHomerun standalone DVR engine Rockon

Hooverdan96:
Fixes #278 .

General information on project
This pull request proposes to add a new rock-on for the following project:

name: HDHomeRun DVR engine from SiliconDust
website: https://www.silicondust.com
description: standalone DVR engine that work in conjunction with SiliconDust's HDHomeRun series of tuners that don't have the DVR built-in. Requires a valid subscription and recent HDHomeRun hardware
Information on docker image
docker image: https://hub.docker.com/repository/docker/jackdock96/hdh_dvr
is an official docker image available for this project?: Not at this time. I created this docker image to self-update with the latest engine and also offer multi-arch images to support Rockstor's direction of installing on ARM architecture systems, so the docker container doesn't have to be maintained in the repository too frequently.
Note: because the dvr engine uses multi-cast the Rockon is currently configured to use the `net=host` option. Unless one implements a `macvlan` or uses tools like `pipework`, multi-cast from LAN to container and vice-versa is not possible. I assume, Rockstor's Rocknets might be able to achieve that, but until we are completely cut over to 4.x on OpenSuse, this is the only option.
Plan is to switch to an official image if/when it becomes available

### Checklist
- [x] Passes [JSONlint](https://jsonlint.com) validation
- [x] Entry added to `root.json` in _alphabetical_ order (for new rock-on only)
- [x] `"description"` object lists and links to the docker image used
- [x] `"description"` object provides information on the image's particularities (advantage over another existing rock-on for the same project, for instance)
- [x] `"website"` object links to project's main website

PR #301, opened 2021-12-02
Add SiliconDust's HdHomerun standalone DVR engine Rockon - restart

### LINKED ISSUE #278: Add HDHomeRun DVR engine Rockon

Hooverdan96:
Fixes #278 . Replaces Pull Request #280 due to messed up fork.

General information on project

name: HDHomeRun DVR engine from SiliconDust
website: https://www.silicondust.com
description: standalone DVR engine that work in conjunction with SiliconDust's HDHomeRun series of tuners that don't have the DVR built-in. Requires a valid subscription and recent HDHomeRun hardware
Information on docker image
docker image: https://hub.docker.com/repository/docker/jackdock96/hdh_dvr
is an official docker image available for this project?: Not at this time. I created this docker image to self-update with the latest engine and also offer multi-arch images to support Rockstor's direction of installing on ARM architecture systems, so the docker container doesn't have to be maintained in the repository too frequently.
Note: because the dvr engine uses multi-cast the Rockon is currently configured to use the net=host option. Unless one implements a macvlan or uses tools like pipework, multi-cast from LAN to container and vice-versa is not possible. I assume, Rockstor's Rocknets might be able to achieve that, but until we are completely cut over to 4.x on OpenSuse, this is the only option.
Plan is to switch to an official image if/when it becomes available

Checklist
 Passes JSONlint validation
 Entry added to root.json in alphabetical order (for new rock-on only)
 "description" object lists and links to the docker image used
 "description" object provides information on the image's particularities (advantage over another existing rock-on for the same project, for instance)
 "website" object links to project's main website


phillxnet (approved):
@Hooverdan96 Yet again my apologies for neglecting this long standing pull request.

I've not real means of testing the actual function of this Rock-on (no hardware and no Web-UI component to the Rock-on) but it installs, stops, starts, and retains it's state on a reboot across both architectures.

Thanks for back-porting your more recent arch text tags. Much appreciated.

So I'll merge and publish and maybe you could then confirm it's function from there on?

Thanks for back-porting your more recent arch text tags. Much appreciated.

HDHomeRun-DVR-Engine-installed

Hooverdan96:
Yes, I will move to the newly merged Rockon shortly, and report back if I run into any issues.

phillxnet:
@Hooverdan96 Cheers.

@FroggyFlox & @Hooverdan96 
PR product published.

Hooverdan96:
@phillxnet I moved over to the "production" and installed fine, survived a Rockstor update related reboot and picked up the configuration I had set for it (recording tasks, serving up existing recordings, etc.).

PR #323, opened 2022-05-24
Update Rockon Readme with Architecture Designation and Format instructions

### LINKED ISSUE #320: Update Rockon readme to advise architecture
In reference to #309 we should probably also add the suggested line with the architecture into the Rockon readme, so new ones will already come with the same tagline on which of the two architectures are supported.

This would also include an update to this section that we had updated via #284 (issue #262), removing the armv7 reference.

Hooverdan96:
Fixes #320.

Add Architecture Designations/explanations to the Readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ If you like to share your app with rest of the Rockstor community, open a pull r
 
 2. Add a comment to your pull request detailing how you've tested it out. More details the better as it will help ensure quality and benefit the whole community!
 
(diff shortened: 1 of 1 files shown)


phillxnet (approved):
@Hooverdan96 Thanks for getting this sorted. Much appreciated.

Thanks also for the added guidance regarding container names.

Not quite sure how the and sign rendering will pan-out within the GitHub Readme rendering but we will see. I'll merge this shortly.

phillxnet:
@FroggyFlox & @Hooverdan96 
PR product in production.

Looks like GitHub rendering worked as intended.

Issue #317, opened 2022-05-21
utorrent image stale


Hooverdan96:
The image used for uTorrent.json was last updated 3 years ago. There is possibly a newer (or official) version that could replace that image?

phillxnet:
@Hooverdan96 & @FroggyFlox I propose then that we simply remove the utorrent Rock-on.

PR #325, opened 2022-05-29
Deprecate haproxy, utorrent & seafile6 Rockons

### LINKED ISSUE #311: haproxy-letsencrypt image not available anymore
The docker image

`illi/docker-haproxy-letsencrypt` is not on docker hub anymore.
The `https://github.com/daniel-illi/docker-haproxy-letsencrypt/tree/rock-on` repo is still there (last updated 6 years ago), but the owner is not hosting it on docker hub anymore. And the original repo it was forked from has also not been active in a number of years.

This Rockon should be deprecated and replaced with an active version. Not sure, whether anybody is currently actively using it and might be interested in creating a new version of this.

Suggestion is to remove this Rockon, as it doesn't work at all. And hope for the best that somebody takes up the challenge to create a replacement.

Hooverdan96:
Fixes #311 and #317.

Since both action require deletion from shared file root.json, proposing to remove both Rockons in the same request.


Hooverdan96:
Looking at the diffs in the root.json, I don't know why it seems I removed AND added the gotify.json again (which I didn't during my editing) ...

phillxnet:
@Hooverdan96 

> Looking at the diffs in the root.json, I don't know why it seems I removed AND added the gotify.json again (which I didn't during my editing) ...

Yes I saw that. It may just be a git/GitHub quirk. Looks like the end result is as expected however.

Hooverdan96:
I also added the seafile6 to the deprecation list. Hope that is ok.

phillxnet (approved):
@Hooverdan96 Thanks for this further Rock-ons update. Much appreciated.

I've tested the product of the pr locally and a Rock-ons 'update' looks to have worked as expected. I'll merger and publish shortly.

Cheers.

phillxnet:
@FroggyFlox & @Hooverdan96 
PR product published to production.

PR #331, opened 2022-06-12
Move scrutiny to official image, altering options accordingly #329

### LINKED ISSUE #329: Scrutiny Rockon not functional currently
Based on Rockstor forum member ITStaff's post (https://forum.rockstor.com/t/disk-unusable-as-pool-member/8403/17), it appears that the Scrutiny Rockon is not working currently.
After checking my install I am experiencing issues as well, the container is running, however it is not functional. After a short investigation I found that this issue has been opened:
https://github.com/linuxserver/docker-scrutiny/issues/22

they are in the process of discussing on how to proceed as scrutiny now needs and InfluxDB instance, which is not (yet) part of the linuxserver.io image.

I propose, that we wait for a little bit until Linuxserver makes a decision on how to proceed (requiring a separate container for the database portion, or incorporating it), before taking any action.

Alternative is to force an older version of linuxserver's image until further notice that does not have that prerequisite (was pointed out in the issue reported to lsio).

phillxnet:
Move from linuxserver.io to official master-omnibus image and remove redundant settings.

Fixes #329
@FroggyFlox & @Hooverdan96 Ready for review. Tests carried out to date to follow in comments.

## Includes:
- Rewording and simplifying instructions.
- Capitalisation fixes.
- Removal of SCRUTINY_API_ENDPOINT option (to simplify).
- Adding COLLECTOR_CRON_SCHEDULE set option. It is often desired to control of when a collection is enacted. This also surfaces that this is only done once per day.
- Update Rock-on json file version number.
- Correct prior mix of tab and spaces.

### Caveats:
The "--privileged" option was found to be required but was not
referenced in the upstream documentation! We already have the two
referenced "--cap-add".

### Checklist


phillxnet:
Tested so far only on amd64 and within KVM with single sata emulated drive. Looks to be working as expected.

## Install config:
quoted-cron-setting

Resulting env:
```
rleap15-3:~ # docker exec scrutiny env
PATH=/opt/scrutiny/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=033c6cd78a1f
COLLECTOR_CRON_SCHEDULE="1 1 * * *"
INFLUXD_CONFIG_PATH=/opt/scrutiny/influxdb
HOME=/root
```
## Installed
scrutiny-installed


## Info panel formatting
scrutiny-rockon-info-panel

## Dashboard
scrutiny-dashboard

## Notes
- No CRON function confirmation as yet.
- Why must we use the --privaledge option as this is not referenced in the upstream docs? Plus we already have both the standard and NvME related --add-cap options!

scrutiny.json
```diff
@@ -28,55 +29,35 @@
...
                     "8080": {
-                        "description": "Port for scrutiny's web interface and API.",
+                        "description": "Port for Scrutiny's web interface and API.",
                         "host_default": 8080,
                         "label": "WebUI HTTP port",
                         "ui": true
                     }
                 },
                 "volumes": {
                     "/config": {
```
AnalogJ, line 39: if this is supposed to be the path within the container, this is incorrect - it should be `/opt/scrutiny/config`

phillxnet: @AnalogJ Thanks, yes I did just leaving that as-was and failed to check. I wondered why the syslog kept apparently indicating the creation of a new scrutiny config. Sorted via a follow-up commit.
Cheers.

Hooverdan96:
I tested it, seems to work. However, I had to clean up the previous Scrutiny share that I had set up for the lsio version. I also used a different port (which mapped just fine). All on amd64.

Small cosmetic suggestions (punctuation/apostrophe removal already applied)
"description": "Choose a Share for Scrutiny's configuration files. **_E.g.:_** create a Share called scrutiny-conf for this purpose alone."

"more_info": "<p><strong>Note:</strong>This image updates **_its_** stats once a day, you may see

For the Cron stuff, you could use a similar line that we've done for Watchtower:
For a handy CRON expression generator see here: [CRON Generator](https://www.freeformatter.com/cron-expression-generator-quartz.html#crongenerator) ... but that's more of a detail, especially if there is a default for the once a day. I just find it hard to read the tooltip and remember what's in there if it contains more than one variable :).

phillxnet:
@Hooverdan96 Thanks for the additional input on this one.

> **Note:**This image updates **_its_** stats once a day, you may see

Sorted (I always seem to miss/create this particular grammatical mistake).

And:

> For the Cron stuff, you could use a similar line that we've done for Watchtower:

Good point: I've opted instead for: https://crontab-generator.com/
as it has only 1 functional session cookie (as apposed to 7 with GA included and a dark pattern cookie consent dialog), is way simpler, and does only the required 5 fields.
I've also popped this in to the description so folks can open ahead of time.

description-with-crontab-helper-link

Hopefully that covers things thus far.

Tested via (showing updated non-linuxserver.io local directory mapping):
corrected-vol-mapping

dev-0.4.13

Now tested on arm64 architecture also (ver dev-0.4.13):
Web-UI appeared but no drives were detected as this system used virtio drives. These do not have smart info available.

phillxnet:
@FroggyFlox  @Hooverdan96 and @AnalogJ 
We do have a little potential fall-out from this one as I think in some settings, a prior image install can cause a Web-UI error during Rock-on update post Rock-on publish indicating that the existing scrutiny Rock-on needs to be uninstalled. This makes sense and is an obvious enough error message but just wanted to make a note here to that end.

I could have done as you have both previously done and taken the re-naming route rather than an in-place edit but it was broken anyway so I think we are OK on this front. I'll go ahead and merge as we have dual arch testing and a fair bit of feedback and tweaks thus far and, as always, we can re-visit as and when.

@AnalogJ Finally I'd like to thank you again for such an excellent project, a really nice fit as a Rockstor Rock-on. And thanks also for correcting my silly directly mistake earlier-on in this pr.

phillxnet:
@Hooverdan96 @FroggyFlox @AnalogJ 
PR product production published.

Hooverdan96:
@phillxnet, thanks for taking this all the way through. I found that when installing the new version, I could retain the "old" data by deleting everything from that share but the <code>scrutiny.db</code> file. Then no error occurs and the continuity of the history seems to be retained. I can post a blurb for that on the forum, if that makes sense.

phillxnet:
@Hooverdan96 Cheers.

Re:

> ... I could retain the "old" data by deleting everything from that share but the `scrutiny.db` file

Nice. Thanks for that. We can add a section to our docs on this Rock-on when they get updated to indicate this transitional move regarding the image switch.

> I can post a blurb for that on the forum, if that makes sense.

Cheers. It could go in the usual Rock-ons updates announcement possibly.

Issue #332, opened 2022-06-19
TP-Link Omada-Controller Rock-On 


janholtzhausen:
Modified the unify rock-on to run mbentley/omada-controller

phillxnet:
@janholtzhausen Hello there.

So could you give some more context / explanation here. I'm assuming you are about to present a pull request?

janholtzhausen:
Hi, thats right, still figuring out the correct procedure :)

janholtzhausen:
I hope it's not too much of a mess..

phillxnet:
@janholtzhausen 
Re:

> I hope it's not too much of a mess..

I wouldn't worry about that. ;)

We do have the following doc entry that may help with procedure:
https://rockstor.com/docs/contribute/contribute_rockons.html

janholtzhausen:
There, I think I checked all the boxes.
The next one will be a cleaner commit, I promise.

Hooverdan96:
@janholtzhausen I can see the Rockon in your forked repo itself, however you would still need to initiate the Pull Request from that branch so it becomes visible over here.

As you do this, quick comment on the description, I think at this point we're stopping to support the armv7 version, as that's still 32-bit and focus on amd64 and arm64 only. So you can take that out before creating the Pull Request.

janholtzhausen:
Is there anything else I need to do?

Hooverdan96:
@janholtzhausen, I don't think so, someone just needs to review it. i will see whether I can do it in the next few days.

janholtzhausen:
There's a couple more services I can contribute, but I'm not sure there's any interest?

Hooverdan96:
@janholtzhausen I would imagine there will be some interest, if you want to list the areas you're thinking about. I have tried to test your Rockon, but apparently I am running into some issues. I've successfully installed it outside of Rockstor, but I seem to be having some issues with my testbed for Rockons that I haven't figured out yet (ran into issues when testing other rockon submissions). Once I can resolve those, I will retest again. The lack of progress on these has more to do with available capacity than lack of interest. So, again thanks for your submission on the TP controller, we'll get it in.

Hooverdan96:
@janholtzhausen, sorry for the delay. I got to install it finally. I left some comments directly in the pull request, if you want to take a look at those.

PR #333, opened 2022-06-20
#332 add project omada controller as rock on

janholtzhausen:
Basics done (I think), please feel free to correct me for future rock-ons :)

### DIFF
--- a/omada-controller.json
+++ b/omada-controller.json
@@ -54,7 +54,7 @@
                         "description": "Choose a LOGS Share for Omada Controller files. Eg: create a Share called omada-controller-logs for this purpose alone.",
                         "label": "EAPController LOGS Storage"
                     }
-                },
+                }
             }
         },
         "description": "Omada Access Point controller.<p>Based on a custom docker image: <a href='https://hub.docker.com/r/mbentley/omada-controller' target='_blank'>https://hub.docker.com/r/mbentley/omada-controller</a>, available for amd64 and arm64 architecture.</p>",
--- a/omada-controller.json
+++ b/omada-controller.json
@@ -43,15 +43,15 @@
                 },
                 "volumes": {
                     "/opt/tplink/EAPController/data": {
-                        "description": "Choose a DATA Share for Omada Controller files. Eg: create a Share called omada-controller for this purpose alone.",
+                        "description": "Choose a DATA Share for Omada Controller files. Eg: create a Share called omada-controller-data for this purpose alone.",
                         "label": "EAPController DATA Storage"
                     },
                     "/opt/tplink/EAPController/work": {
-                        "description": "Choose a WORK Share for Omada Controller files. Eg: create a Share called omada-controller for this purpose alone.",
+                        "description": "Choose a WORK Share for Omada Controller files. Eg: create a Share called omada-controller-work for this purpose alone.",
                         "label": "EAPController WORK Storage"
                     },
                     "/opt/tplink/EAPController/logs": {
-                        "description": "Choose a LOGS Share for Omada Controller files. Eg: create a Share called omada-controller for this purpose alone.",
+                        "description": "Choose a LOGS Share for Omada Controller files. Eg: create a Share called omada-controller-logs for this purpose alone.",
                         "label": "EAPController LOGS Storage"
                     }
                 },
--- a/omada-controller.json
+++ b/omada-controller.json
@@ -55,13 +55,6 @@
                         "label": "EAPController LOGS Storage"
                     }
                 },
-                "environment": {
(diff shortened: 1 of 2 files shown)


Hooverdan96:
Fixes #332

omada-controller.json
```diff
@@ -0,0 +1,76 @@
...
+                        "label": "EAPController LOGS Storage"
+                    }
+                },
+                "environment": {
+                    "TZ": {
+                    "description": "Timezone for the controller.",
+                    "label": "TZ",
+                    "index": 1	
+                    }
+                }
```
Hooverdan96, line 64: The timezone mapping is not really required. The Rockon helper already maps those to the TZ of the Rockstor appliance
```suggestion

```

Hooverdan96:
@janholtzhausen, I believe the Timzone mapping is not required at this time, as Rockstor always includes it as part of the Docker command it issues, tied to the appliance timezone.

I noticed, that the container changes the permissions on the share to the default process ID/group (508/508). Would it be better to expose the PUID/PGID parameters in the configuration, so one has better control over on how the shares are managed, as opposed to the Rockon managing the share (and one then not necessarily having access to it anymore, unless using the root user)?

Since I don't have relevant TP-Link controllers, I can't really test, but does this setup require the `net=host` setting? I wasn't sure, and we are striving to not have to use it, but just wanted to make sure, that it will work without it.

omada-controller.json
```diff
@@ -0,0 +1,69 @@
...
+                    },
+                    "/opt/tplink/EAPController/work": {
+                        "description": "Choose a WORK Share for Omada Controller files. Eg: create a Share called omada-controller-work for this purpose alone.",
+                        "label": "EAPController WORK Storage"
+                    },
+                    "/opt/tplink/EAPController/logs": {
+                        "description": "Choose a LOGS Share for Omada Controller files. Eg: create a Share called omada-controller-logs for this purpose alone.",
+                        "label": "EAPController LOGS Storage"
+                    }
+                },
```
Hooverdan96, line 57: @janholtzhausen. My apologies, as part of the changes requested I overlooked that there was a superfluous comma hanging around. Could you remove that one (it then passes the JSON lint test, too).
```suggestion
                }
```

Hooverdan96:
@janholtzhausen thank you! I'll test one more time tomorrow and then it should be good to go from my perspective.

janholtzhausen:
Hi,
I think it;s necessary, since the protocols are not too different from the Ubiquity kit.

Hooverdan96:
@janholtzhausen, thinking about it some more, if the current version you pushed here works for you (meaning running the container within the Rockon framework and connecting to the Omada controller is not a problem for you) then it does not seem to be necessary. And would be what we'd prefer anyway, since it makes port management simpler between Rockons.
Otherwise, if you could update the json file to contain the parameter variable?
image

kanecko:
I'm amidst creating my own Rockon, because I have discovered this PR too late.
What's the progress on this PR? Was it canceled?

phillxnet:
@kanecko Hello there. Re:

> What's the progress on this PR? Was it canceled?

I'm not actually sure myself. From the pr history it looks like @Hooverdan96 was exploring the need for the --net=host option as we only like to accept that if it's absolutely necessary, given it gives the docker image full network access. There seems to have been a question as to it's need or otherwise.

A month later @janholtzhausen looks to have closed the pull request. It could be that they lost interest in the submission or no longer had access to a rockstor system to prove any changes.

If you fancy submitting your own then do go ahead, we are limited in the number of folks who review these Rock-on submissions however so a bit of patience may be required unfortunately. Just take a looks at our other Rock-ons and some more recent accepted (merged) Rock-on submissions to see the type and format of information as that should drastically speed up the review process. The actual publishing is relatively trivial however.

kanecko:
Sure, why not 😊
Will create a new PR at some point.

janholtzhausen:
Afraid my Rockstor solution became redundant.

 

I was pretty sure that host level access was still required, just because of an element of dynamic port usage for device discovery (pretty much the same situation as the Ubiqity controller rock-on, which this was shamelessly copied from…)

 

I can spin up a virtual rockstor to re-test, but I’m afraid a nearly yearlong delay between PR and acceptance may be too long for me to keep a dev instance alive in the lab.

 

 

Regards

Jan

kanecko:
Thanks for clarifying!
I have my own Omada Rockon up and running without the `--net=host` option. At least the WebUI works.
I will set up my Omada devices next, to see, if it still works without it.

kanecko:
Everything is working fine.
submitted PR #341

Issue #328, opened 2022-06-06
A tailscale Rock-on would be nice


phillxnet:
There is the following official Tailscale docker image:
https://hub.docker.com/r/tailscale/tailscale

But it's not looking like we need it to currently. Just noting here in case we get any takers or the work to adapt to this rock-on within the rockstor-core takes anyone's fancy. Or if anyone can fathom a way we can get this official docker image in as-is. It could go in with command line caveats/post install setup but that's really not in the Rock-on remit really.

phillxnet:
This issue is awaiting the final testing/review/and merge of #336 to help clear the way before adding yet more Rock-ons.

phillxnet:
Given the requirement of the official docker image of Tailscale, the hightly preferred option given the security and network concernes here, is it thought that we would need custom custom config & start options such as are employed (in config and install) for our current open-vpn and owncloud rock-ons here:

https://github.com/rockstor/rockstor-core/blob/master/src/rockstor/storageadmin/views/rockon_helpers.py

Given that such treatment would lock any Tailscale Rock-on here to only future Rockstor versions it is currently considered that the development time may well be better spent on intergrating Tailscale 'properly' as a service, which is what it runs as anyway (using a go binary). Repo addition logistic is a related concern here. Expect a newer issue in core and the likely closing of this issue if there is not further follow up for some time.

Stick point in Rock-on development was the passing of an authentication key:
From: https://hub.docker.com/r/tailscale/tailscale

We require the input of a pre-created via tailscale dash settings-key tailnet auth key and for this to be passed to the docker container via:
```
docker exec tailscaled tailscale up --authkey=$KEY
```

Hence the suggestion that our current setup could do this only via custom_config

      (optional)"custom_config": <custom configuration object that a special install handler of this Rock-on expects>

entry and dedicated rockon.name.lower()_start and/or install code adition.

Noting here just in case this is re-visited. But the current side channel consensus with admittedly only parsing consideration is that Tailscale is sufficiently inportant to adopt as a core service where we can vastly improve our intergration/compatibility in-comparison to supporting only from a Rock-on where we have already run into an implementation issue.

My apologies if I've missed an obviously implementation trick here.
[...]

PR #338, opened 2022-09-27
Tailscale Rock-on incomplete #328

phillxnet:
Currently incomplete and awaiting upstream modifications to the official docker run.sh to enable web based initialisation, and re-auth when required.

It is proposed that in-time we focus on adding a Tailscale service 'proper' given this technologies value, and mind share, for future capabilities regarding remote access, an oft requested config/capability. However a Rock-on may serve well in the mean time given our limited developer resources.

- [ ] Add status command guide to info section: "docker exec tailscaled tailscale --socket=/tmp/tailscaled.sock status" as this would be usefull for diagnostic purposes.

### DIFF
--- a/root.json
+++ b/root.json
@@ -60,6 +60,7 @@
     "SmokePing": "smokeping.json",
     "Sonarr": "sonarr.json",
     "Syncthing": "syncthing.json",
+    "Tailscale": "tailscale.json",
     "Tautulli": "tautulli.json",
     "TeamSpeak3": "teamspeak3.json",
     "Transmission with OpenVPN": "transmission-with-openvpn.json",
--- a/tailscale.json
+++ b/tailscale.json
@@ -1,10 +1,10 @@
 {
     "Tailscale": {
         "containers": {
-            "launch_order": 1,
             "tailscaled": {
                 "image": "tailscale/tailscale",
                 "tag": "stable",
+                "launch_order": 1,
                 "opts": [
                     [
                         "--net=host",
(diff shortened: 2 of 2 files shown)


phillxnet:
Adding as draft in case folks find this usefull and to aid further discussion in case I've missed an easy fix in this case. Please see issue text for the blocker experienced on further development of this Rock-on.

phillxnet:
Installed view so far:

Rock-on_installed-part-functional

phillxnet:
After the move to use only the build-in webui that re-directs to the auth url required we have a more refined on-boarding:

post-webui-config-option-add-pre-install

With only one port and one state dir (Share) to assign and we are done bar the WebUI auth bit.

Hooverdan96:
@phillxnet, I just happened to look at this today, and it seems that your pull request to the tailscale repo has become somewhat obsolete, as they did away with the `run.sh` and replaced it with a go-based implementation with this commit:
https://github.com/tailscale/tailscale/commit/76904b82e7b474b14e58bad9f2668409b816fc43
Sadly, your proposal was still not included there.

phillxnet:
Closing as the lack of upstream flexibility renders this effort mute. We will have to take a differing tack no this front I think.

phillxnet:
No planning openSUSE level intergration avoiding any limitations regarding the docker approach. Noting here an encouraging interaction with upstream on our prior efforts to contribute there: 

https://github.com/tailscale/tailscale/issues/5799#issuecomment-1575192573

details Unraid support added to their codebase. We likely need to follow though and contribute something similar to ensure an easier path for us in the future.

@Hooverdan96 the above comment and linked (now merged) pull requests may be of some interest.

Hooverdan96:
Reading more on tailscale, I would tend to agree with the openSUSE integration. My assumption would be, if a Rockstor owner will be using it in a corporate setting, they will go for the paid license setup, and I would further assume that a corporate IT environment will prefer a OS level integration vs. managing a pure docker solution?
In any case, that seems to be the better approach considering the upstream convo you've been having and the approaches taken for other appliances.

PR #342, opened 2023-03-04
Pi-Hole Rockon update to enable dhcp server

### LINKED ISSUE #305: Pi-hole not exposing DHCP server ports
Pi-hole's DHCP server option is not working, I believe because the DHCP server port (68) is not exposed.

Hooverdan96:
Fixes #305. 
Update Rockon with net=host option to easily enable Pi-hole use as dhcp server.
Remove superfluous port mappings due to net=host option being active.
Cosmetic description adjustments and to highlight net=host option being active

This pull request proposes to update the pi-hole Rockon.

- official docker image: (https://hub.docker.com/r/pihole/pihole) - unchanged from previous Rockon version

### Checklist

### DIFF
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -46,7 +46,7 @@
 				},
 				"environment": {
 					"FTLCONF_LOCAL_IPV4": {
-						"description": "Enter ip4v adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
+						"description": "Enter IPv4 adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
 						"label": "Rockstor IP",
 						"index": 1
 					},
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -51,7 +51,7 @@
 						"index": 1
 					},
 					"WEB_PORT": {
-						"description": "Enter the Port Name for the WebUI. Web-port. Recommended: 80. This port has to be the same as the Rockon UI Link port, configured earlier. This ensures that the admin pages for pi-hole are accessible via Web Browser",
+						"description": "Enter the WebUI Web-port number. Recommended: 80. This port has to be the same as the Rockon UI Link port, configured earlier. This ensures that the admin pages for pi-hole are accessible via Web Browser",
 						"label": "Web-Port",
 						"index": 2
 					},
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -46,7 +46,7 @@
 				},
 				"environment": {
 					"FTLCONF_LOCAL_IPV4": {
-						"description": "Enter IPv4 adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
+						"description": "Enter the IPv4 address of the Rockstor host. If not specified, it will default to the internal docker IP and will not work.",
 						"label": "Rockstor IP",
 						"index": 1
 					},
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -46,7 +46,7 @@
 				},
 				"environment": {
 					"FTLCONF_LOCAL_IPV4": {
-						"description": "Enter IP-adress of rockstor server. If not specified it will default to internal docker IP and it will not work.",
+						"description": "Enter ip4v adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
 						"label": "Rockstor IP",
 						"index": 1
 					},
@@ -66,6 +66,9 @@
 		"description": "DNS Server that acts as a black hole for Internet advertisements.<p>Based on the official docker image: <a href='https://hub.docker.com/r/pihole/pihole' target='_blank'>https://hub.docker.com/r/pihole/pihole</a>, available for amd64 and arm64 architecture.</p>",
 		"more_info": "<h4>PI-HOLE™: A BLACK HOLE FOR INTERNET ADVERTISEMENTS</h4><p><b>Admin page</b></p><p>To access admin interface go to URL: http://[SERVERIP]:[Web-Port]/Admin</p><p><b>Block Over 100,000 Ad-serving Domains</b></p><p>Known ad-serving domains are pulled from third party sources and compiled into one list.</p><p><b>Block Advertisements On Any Device</b></p><p>Network-level blocking allows any device to block ads, regardless of hardware or OS.</p><p><b>Improve Overall Network Performance</b></p><p>Since ads are blocked before they are downloaded, your network will perform better.</p><p>Note: this Rockon uses the net=host option to allow for use of the dhcp functionality.",
 		"volume_add_support": false,
+        	"ui": {
+            	"slug": "admin"
+        	},		
 		"website": "https://pi-hole.net/",
 		"version": "2.5"
 	}
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -1,72 +1,72 @@
 {
-    "Pi-Hole": {
-        "containers": {
-            "pi-hole": {
-                "image": "pihole/pihole",
-                "opts": [
-                    [
-                        "--cap-add",
-                        "NET_ADMIN"
-                    ],
-                    [
-			"--dns",
-			"127.0.0.1"
-                    ],
-		    [
-			"--dns",
-			"8.8.8.8"
-		    ],
-		    [
-			"-e",
-			"IPv6=False"
-		    ]
-                ],
-                "launch_order": 1,
-                "ports": {
-                    "53": {
-                        "description": "DNS port. Required: 53",
-                        "host_default": 53,
-                        "label": "DNS-Port",
-                        "ui": false
-                    },
+	"Pi-Hole": {
+		"containers": {
+			"pi-hole": {
+				"image": "pihole/pihole",
+				"opts": [
+					[
+						"--cap-add",
+						"NET_ADMIN"
+					],
+					[
+						"--dns",
+						"127.0.0.1"
+					],
+					[
+						"--dns",
+						"8.8.8.8"
+					],
+					[
+						"--net",
+						"host"
+					],
+					[
+						"-e",
+						"IPv6=False"
+					]
+				],
+				"launch_order": 1,
+				"ports": {
                     "80": {
-                        "description": "Web-port. Recommended: 80. If other port than 80 is used some blocked sites will not show correctly.",
+                        "description": "Rockon UI Link Port. Recommended: 80. This port has to be the same as the subsequent Web-port later in the configuration, so the admin pages can be accessed from within the Rockon page.",
                         "host_default": 80,
-                        "label": "Web-Port",
+                        "label": "Rockon_Web-Port",
                         "ui": true
                     }
                 },
-                "volumes": {
-                    "/etc/pihole": {
-                        "description": "Choose a share for Pi-Hole configuration. Eg: create a Share called pihole-config for this purpose alone.",
-                        "label": "Pi-Hole config"
-                    },
-		    "/etc/dnsmasq.d": {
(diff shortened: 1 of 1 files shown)


Hooverdan96:
@phillxnet, @FroggyFlox here's the PR as discussed. I have done install/a bit of Web Admin administration (did not test the dhcp server option, since I couldn't mess up my network)/uninstall. All worked without error on my machine

pi-hole.json
```diff
@@ -1,72 +1,72 @@
...
+					}
+				},
+				"environment": {
+					"FTLCONF_LOCAL_IPV4": {
+						"description": "Enter IP-adress of rockstor server. If not specified it will default to internal docker IP and it will not work.",
+						"label": "Rockstor IP",
+						"index": 1
+					},
+					"WEB_PORT": {
+						"description": "Enter the Port Name for the WebUI. Web-port. Recommended: 80. This port has to be the same as the Rockon UI Link port, configured earlier. This ensures that the admin pages for pi-hole are accessible via Web Browser",
```
phillxnet, line 54:
```suggestion
						"description": "Enter the WebUI Web-port number. Recommended: 80. This port has to be the same as the Rockon UI Link port, configured earlier. This ensures that the admin pages for pi-hole are accessible via Web Browser",
```

phillxnet (changes requested):
@Hooverdan96 Thanks for seeing to this change. I'm currently doing a test install but the suggested change reads better to me. Or am I miss-reading what is intended here.

pi-hole.json
```diff
@@ -1,72 +1,72 @@
...
+						"label": "Pi-Hole config"
+					},
+					"/etc/dnsmasq.d": {
+						"description": "Choose a share for dnsmasq configuration. E.g.: create a Share called dnsmasq-config for this purpose alone.",
+						"label": "dnsmasq config"
+					}
+				},
+				"environment": {
+					"FTLCONF_LOCAL_IPV4": {
+						"description": "Enter IP-adress of rockstor server. If not specified it will default to internal docker IP and it will not work.",
```
phillxnet, line 49: @Hooverdan96 & @FroggyFlox should the user facing text here specify an IPv4 address? The env name would suggest this.

phillxnet:
When installing thus:

Share | dnsmasq-config | /etc/dnsmasq.d
Share | pihole-config | /etc/pihole
Port | 80 | 80
Env | 192.168.2.196 | FTLCONF_LOCAL_IPV4
Env | 80 | WEB_PORT
Env | anotrealpass | WEBPASSWORD

I get a "403 Forbidden" on the Web-UI button. Am I doing it wrong?

pi-hole.json
Hooverdan96: yep, that is of course the correct way, don't know what I was thinking

pi-hole.json
Hooverdan96: Yes, that's how I have interpreted it from the previous version of this Rockon. I can specify that out

Hooverdan96:
That is interesting indeed. this is probably the same behavior as we had in the previous version.
I guess, I made the mistake of not using the recommended port for the installation, but the suggested one by Rockstor (in my case 83). So maybe the 'recommended' port option should not be stated ...
```
[i] WARNING: running in host network mode forces lighttpd's bind address to $FTLCONF_LOCAL_IPV4 (192.168.2.196).
This behaviour is deprecated and will be removed in a future version. If your installation depends on a custom bind address (not 0.0.0.0) you should set the $WEB_BIND_ADDR environment variable to the desired value.
```
I have to read up on that, because I followed the docker variable descriptions (and what's deprecated and all that) on the docker git ...

Hooverdan96:
This is what I found in reference to the warning messages. I guess, it's still ok to use it, despite the warning.
image
and
image

phillxnet:
I have just tested changing the DNS entry on a 'smart' TV to use my above configured Pi-Hole. It does appear to be blocking many calls from that TV, at least as reported by the Pi-Hole Dashboard. This TV is on the same network so I could just modify the DNS it used to that of the Rockstor machine. That looks to be a win I think.

Hooverdan96:
I did the additional test of using the Rockstor host name, after installation:
`https://rockleaper/` (which takes me to the Rockstor WebUI, the variation with `rockleaper.lan` does not work at all in my network)
the pi-Hole UI link:
`http://rockleaper:80/` puts me right here: `http://rockleaper/admin/login.php`

so, for me both logging in through the actual Rockstor IP or the virtual host name and then starting the Rockon UI for pi-hole seems to work.

Here's an excerpt of the docker logs for during the startup:
```
  [i] Added ENV to php:
                    "TZ" => "",
                    "PIHOLE_DOCKER_TAG" => "",
                    "PHP_ERROR_LOG" => "/var/log/lighttpd/error-pihole.log",
                    "CORS_HOSTS" => "",
                    "VIRTUAL_HOST" => "rockleaper",
  [i] Using IPv4
  [i] WARNING: running in host network mode forces lighttpd's bind address to $FTLCONF_LOCAL_IPV4 (172.28.128.9).
  [i] This behaviour is deprecated and will be removed in a future version. If your installation depends on a custom bind address (not 0.0.0.0) you should set the $WEB_BIND_ADDR environment variable to the desired value.
  [i] Preexisting ad list /etc/pihole/adlists.list detected (exiting setup_blocklists early)
  [i] Existing DNS servers detected in setupVars.conf. Leaving them alone
  [i] Applying pihole-FTL.conf setting LOCAL_IPV4=172.28.128.9
  [i] FTL binding to default interface: eth0
  [i] Enabling Query Logging
  [i] Testing lighttpd config: Syntax OK
  [i] All config checks passed, cleared for startup ...
  [i] Docker start setup complete
  [i] pihole-FTL (no-daemon) will be started as pihole
```
so, very similar to what you were seeing.

FroggyFlox:
> I have just tested changing the DNS entry on a 'smart' TV to use my above configured Pi-Hole. It does appear to be blocking many calls from that TV, at least as reported by the Pi-Hole Dashboard.

Oh I know that feeling... I've been in the same situation before and since disconnected it :-)

We seem to have a few things here:

>  and the Pi-Hole Web-UI link is: http://rleap15-4.lan/
That is because Rockstor builds the UI link as follows:
- use `http` or `https` based on the Rock-on definition
- then append Rockstor's hostname as seen by the JS `window.location.hostname` (so that includes the `.lan`).
- then append the port number that has the `ui: true` set in the Rock-on definition (if any)
- then append the slug if defined

Given your Rockstor hostname has the `.lan`, it makes sense that it was built that way.

Then:
> But if I visit (with the above install settings of the recommended port 80 where indicated):
> http://rleap15-4/
> I.e. not ".lan" I get redirected to: http://rleap15-4/admin/login.php

That is related to the `VIRTUAL_HOST` env (set by default to the machine's hostname `${HOSTNAME}`):
> What your web server 'virtual host' is, accessing admin through this Hostname/IP allows you to make changes to the whitelist / blacklists in addition to the default 'http://pi.hole/admin/' address
Sooo... my guess is that `echo $HOSTNAME` returns `rleap15-4` and not `rleap15-4.lan` for you?

FroggyFlox:
@phillxnet ... would using a more exotic port number than 80 make the UI button work for you, by any chance (as the Port and WEB_PORT settings)? Could Rockstor's nginx config mess with the use of the Port 80 here?

phillxnet:
@FroggyFlox Re:

> Sooo... my guess is that `echo $HOSTNAME` returns `rleap15-4` and not `rleap15-4.lan` for you?

> rleap15-4:~ # echo $HOSTNAME
> rleap15-4

Indeed :). What I need is a Pi-Hole to manage this network :). Or to put in the full name (.lan included) during Web-UI setup I suspect. I'll experiment as I can already resolve from a client with either.

> @phillxnet ... would using a more exotic port number than 80 make the UI button work for you, by any chance (as the Port and WEB_PORT settings)?

I don't know if 88 counts but I did try that with identical effect. So changed back given my unfamiliarity with Pi-Hole and the knowledge already backed in via the recommendations to use 80. I should have mentioned this test earlier. My apologies.

> Could Rockstor's nginx config mess with the use of the Port 80 here?

That was my first thought and why I then moved to 88. But I think we basically now ignore port 80.

Plus re:

> Oh I know that feeling... I've been in the same situation before and since disconnected it :-)

The TV is the only client set to use this Pi-Hole Rock-on and I've seen the following graph since that test began:
(pop-up non representative (smallest number of requests) but gave a better view of the graph.

Blocked-Queries

@Hooverdan96 Thanks for the additional test, explanations. I'll retest using the new 
'Text Corrections & adding slug specification' commit likely tomorrow now.

If all is well I think we are better off with this version than we were with the last, so likely we can publish shortly there after.

pi-hole.json
```diff
@@ -46,7 +46,7 @@
 				},
 				"environment": {
 					"FTLCONF_LOCAL_IPV4": {
-						"description": "Enter IP-adress of rockstor server. If not specified it will default to internal docker IP and it will not work.",
+						"description": "Enter ip4v adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
```
phillxnet, line 49: Shouldn't that read IPv4

phillxnet (approved):

phillxnet:
So as I see it we just need one more test/review of this re the slug addition.

pi-hole.json
Hooverdan96: I will correct that and push it.

pi-hole.json
```diff
@@ -1,72 +1,75 @@
...
+						"label": "Pi-Hole config"
+					},
+					"/etc/dnsmasq.d": {
+						"description": "Choose a share for dnsmasq configuration. E.g.: create a Share called dnsmasq-config for this purpose alone.",
+						"label": "dnsmasq config"
+					}
+				},
+				"environment": {
+					"FTLCONF_LOCAL_IPV4": {
+						"description": "Enter IPv4 adress of rockstor server. If not specified, it will default to internal docker IP and it will not work.",
```
phillxnet, line 49:
```suggestion
						"description": "Enter the IPv4 address of the Rockstor host. If not specified, it will default to the internal docker IP and will not work.",
```

phillxnet: A minor niggle typo/text change. Not important but if we get to do this in this pr then fine. Otherwise we can follow up after some field testing as it would be great to publish as-is anyway. Given the improvements thus far.

phillxnet (changes requested):
@Hooverdan96 This is all great work and a relief to now have this working as folks might generally expect.

And my Web-UI button now works when using the suggested settings throught-out.

I've made one last spelling and additional "the" entries in user text. Lets see if we can get them in and get this out as it's a major improvement on what we had.

My existing config from an earlier review was also picked up just dandy and having the Web-UI button work this time is nice.

phillxnet:
It looks like the /admin slug has sorted my issue re the "Pi-Hole UI" button link:

With what we have here to date (slug included) the button works.

https://rleap15-4.lan/ Rockstor Web-UI working
http://rleap15-4.lan/admin "Pi-Hole UI" installed Rock-on button working
http://rleap15-4.lan/ (by-hand) 403 Forbidden (as before on the button without the recent slug entry)
http://rleap15-4/ (by-hand) as before also works and redirects to: http://rleap15-4/admin/login.php

phillxnet (approved):

phillxnet:
@Hooverdan96 Thanks for sticking with this one. And my apologies for the multi-wave of reviews. I'm really not familiar with Pi-Hole so it took me a bit to review properly, and there ended up being distractions along the way - who would have thought :) . But this now looks to be a major improvement on what we had. And looks to function as intended/expected. Much appreciated.

@FroggyFlox Thanks for your help here and on the forum: as always - all good, and much appreciated.

I'll merge and note here once the repo is copied to the production server.

phillxnet:

PR #367, opened 2024-03-10
Update Handbrake.json

### LINKED ISSUE #366: Handbrake Rockon in need of maintenance/deletion
Thanks to forum member GoreMaker for highlighting this issue. Our current mapping for the image used:

https://hub.docker.com/r/jlesage/handbrake
https://github.com/jlesage/docker-handbrake

are likely in need of updating. If no PR is forthcomming it is proposed that this Rock-on is deleted.

- Original addition PR: https://github.com/rockstor/rockon-registry/pull/210
- Subsequent update: https://github.com/rockstor/rockon-registry/pull/213
- Further modifications: https://github.com/rockstor/rockon-registry/pull/270

---

Forum reference: https://forum.rockstor.com/t/handbrake-rockon-required-parameters-seem-wrong/9332

Hooverdan96:
- add optical drive
- change storage label
- clean up comments

Fixes #366  .

### Checklist

### DIFF
--- a/handbrake.json
+++ b/handbrake.json
@@ -1,16 +1,16 @@
 {
-    "HandBrake_test": {
+    "HandBrake": {
         "containers": {
             "handbrake": {
                 "image": "jlesage/handbrake",
                 "launch_order": 1,
                 "devices": {
                     "quicksync": {
-                        "description": "Optional: path to hardware transcoding device for Intel Quick Sync (<code>/dev/dri</code>). Leave blank if not needed.",
+                        "description": "Optional: path to hardware transcoding device for Intel Quick Sync (/dev/dri). Leave blank if not needed.",
                         "label": "Intel Quick Sync device"
                     },
                     "opticaldrive": {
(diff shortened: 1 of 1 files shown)


Hooverdan96:
@phillxnet, @FroggyFlox, I have now cleaned up the original draft PR (e.g. removing html tags from the info texts, and adjusting the 'test' naming so it wouldn't conflict with the existing Rockon during testing).

If/when you get a chance to review, please kindly do so.

phillxnet (approved):
@Hooverdan96 Thanks for seeing to these updates. As always much appreciated.

I only did a basic install to note the format etc. 

I'm assuming the following was intended:
formatting-re-format-preset


My test install used the following settings:
handbrake-installed-config

The initial Web-UI was as follows (cut in height):
Initial-Handbrake-Web-UI

I did notice that the above Web-UI did not indicate the "Very Fast" as pre-selected; but "Fast ...".

No showstoppers from what I can tell so merging. Thanks again for your Rock-on maintenance: always much appreciated.

phillxnet:

Hooverdan96:
@phillxnet, on the above pieces, yes, that was intended, since we don't have an "optional/use default" setup for those parameters yet, so they're still required or installation won't proceed. And, if I remember correctly during the initial conversation on Handbrake some years ago, I think the main user had intended on being able to change the defaults ...

I think there will be a few Rockons that will benefit from the ability to make more parameters options (and possibly with a default value behind them, so even if they're mandatory they would then be pre-populated.

PR #372, opened 2024-04-04
correct GID config option description

unguul:
correct GID config option description following the initial (probably) copy&paste;

Fixes a tiny typo;

This pull request proposes to correct a tiny typo which might cause confusion.

n/a

### Checklist
- [ ] `"website"` object links to project's main website

### DIFF
--- a/duplicati-lsio.json
+++ b/duplicati-lsio.json
@@ -27,7 +27,7 @@
                   "index": 1
                },
                "PGID": {
-                  "description": "Enter a valid UID of an existing user with permission to shares.",
+                  "description": "Enter a valid GID of an existing group with permission to shares.",
                   "label": "PGID",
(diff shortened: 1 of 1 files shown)


Hooverdan96:
@unguul thanks for noticing that. @phillxnet or @FroggyFlox . since this one is truly just a change in a few characters in the description, can you merge this when you get a chance? I agree that this could be confusing.

phillxnet (approved):
@unguul Thanks for seeing to this, much appreciated.

I'm a little surprised this hasn't been noticed before hand !! But here we are :) .

Looks good, and I'll look to publishing shortly.
Once published I'll update this pull request accordingly.

phillxnet:
Corrected-text-re-GUID

PR #393, opened 2024-11-16
Paperless ngx

BrokenOnedroid:

Fixes # .

- name: paperless-ngx
- website: https://docs.paperless-ngx.com/
- description: DMS. I'm checking if i want to migrate from OMV to rockstor.  Adding needed container. 

- docker image: https://github.com/paperless-ngx/paperless-ngx

### Checklist

### DIFF
--- a/paperless-ngx.json
+++ b/paperless-ngx.json
@@ -1,0 +1,101 @@
+{
+    "Paperless-ngx": {
+        "containers": {
+            "paperless": {
+                "image": "ghcr.io/paperless-ngx/paperless-ngx",
+                "launch_order": 3,
+                "ports": {
+                    "8000": {
+                        "description": "Port for web interface. Suggested default: 8420.",
+                        "host_default": 8420,
+                        "label": "WebUI port",
+                        "protocol": "tcp",
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/data": {
+                        "description": "Choose a Share for data file storage.",
+                        "label": "Config paperless Data - Storage"
+                    },
+                    "/media": {
+                        "description": "Choose a Share for media storage.",
+                        "label": "Config paperless Media - Storage"
+                    },
+                    "/export": {
+                        "description": "Choose a Share for exported documents.",
+                        "label": "Config paperless Export - Storage"
+                    },
+                    "/consume": {
+                        "description": "Choose a Share for importing documents.",
+                        "label": "Config paperless Import - Storage"
+                    }
+                },
+                "environment": {
+                    "PAPERLESS_REDIS": {
+                        "description": "Set to redis://broker:6379.Redis connection URL. Do not change unless you know what you are doing.",
+                        "value": "redis://broker:6379",
+                        "label": "Redis URL"
+                    },
+                    "PAPERLESS_DBHOST": {
+                        "description": "Set to db. Hostname of the PostgreSQL database container.",
+                        "value": "db",
+                        "label": "Database Host"
+                    },
+                    "PAPERLESS_TIME_ZONE": {
+                        "description": "Enter timezone",
+                        "value": "Europe/Berlin",
+                        "label": "Timezone"
+                    }
+                }
+            },
+            "broker": {
+                "image": "docker.io/library/redis",
+                "tag": "7",
+                "launch_order": 1,
+                "volumes": {
+                    "/data": {
+                        "description": "Choose a Share for Redis data.",
+                        "label": "Redis Data Share"
+                    }
+                }
+            },
+            "db": {
+                "image": "docker.io/library/postgres",
+                "tag": "16",
+                "launch_order": 2,
+                "volumes": {
+                    "/var/lib/postgresql/data": {
+                        "description": "Choose a Share for PostgreSQL database storage.",
+                        "label": "PostgreSQL Storage"
+                    }
+                },
+                "environment": {
+                    "POSTGRES_DB": {
+                        "description": "Enter the database name for Paperless.",
+                        "label": "Database Name",
+                        "index": 1
+                    },
+                    "POSTGRES_USER": {
+                        "description": "Enter the username for the PostgreSQL database.",
+                        "label": "Database Username",
+                        "index": 2
+                    },
+                    "POSTGRES_PASSWORD": {
+                        "description": "Enter the password for the PostgreSQL database user.",
+                        "label": "Database Password",
+                        "index": 3
+                    }
+                }
+            }
+        },
+        "description": "Paperless DMS - Private & Secure document management system. <p>Based on a Docker image: <a href='https://github.com/paperless-ngx/paperless-ngx/tree/main'>GitHub Repository</a>. Available for amd64 and arm64 architectures.</p> <p> Create User like shown here (#7): <a href='https://docs.paperless-ngx.com/setup/#docker_hub'>https://docs.paperless-ngx.com/setup/#docker_hub</a> </p>",
+        "ui": {
+            "https": false,
+            "slug": ""
+        },
+        "volume_add_support": true,
+        "website": "https://docs.paperless-ngx.com/",
+        "version": "1.2"
+    }
+}
--- a/root.json
+++ b/root.json
@@ -77,5 +77,6 @@
     "WireGuard Easy": "WireGuardEasy.json",
     "YouTrack official": "youtrack-official.json",
     "Zabbix-XXL": "zabbix-xxl.json",
-    "ZeroNet": "ZeroNet.json"
+    "ZeroNet": "ZeroNet.json",
+    "Paperless-ngx": "paperless-ngx.json"
 }


Hooverdan96:
@BrokenOnedroid thanks for this PR. This makes a lot of sense to add to the Rockon repository.

I think there are a couple of cosmetic things that could be adjusted for the install wizard, but let's get to that later. One thing I would like to ask you is if you could move the new entry up in the `root.json` so it fits into the alphabetical order of the Rockons listed.

I am testing the install right now, however I am running into an issue, maybe you can clarify.

Installation Wizard outputs:
Listing in the Rockon directory (I manually added it using the `rockons-metastore` for now.
image

Shares:
image


WebUI Port:
image


Environment variables:
image


Summary before install:
image


Resulting installed instance:
image


However, when I navigate to the WebUI, it refuses to connect:
image

When checking with `docker ps`, I can see that while postgres and redis are running continuously, the paperless container restarts every approx. 35 seconds or so:
image

I suspect, the reason for that restart is here:
`docker logs paperless`
```
Paperless-ngx docker container starting...
Creating directory scratch directory /tmp/paperless
Adjusting permissions of paperless files. This may take a while.
Waiting for PostgreSQL to start...
Connected to PostgreSQL
Waiting for Redis...
Redis ping #0 failed.
Error: Error -5 connecting to broker:6379. No address associated with hostname..
Waiting 5s
Redis ping #1 failed.
Error: Error -5 connecting to broker:6379. No address associated with hostname..
Waiting 5s
Redis ping #2 failed.
Error: Error -5 connecting to broker:6379. No address associated with hostname..
Waiting 5s
Redis ping #3 failed.
Error: Error -5 connecting to broker:6379. No address associated with hostname..
Waiting 5s
Redis ping #4 failed.
Error: Error -5 connecting to broker:6379. No address associated with hostname..
Waiting 5s
Failed to connect to redis using environment variable PAPERLESS_REDIS.
Paperless-ngx docker container starting...
```
The redis container seems to be running just fine:
```
rockwurst: # docker logs broker
1:C 17 Nov 2024 11:53:47.607 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
1:C 17 Nov 2024 11:53:47.608 * oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 17 Nov 2024 11:53:47.608 * Redis version=7.4.1, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 17 Nov 2024 11:53:47.609 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
1:M 17 Nov 2024 11:53:47.610 * monotonic clock: POSIX clock_gettime
1:M 17 Nov 2024 11:53:47.611 * Running mode=standalone, port=6379.
1:M 17 Nov 2024 11:53:47.611 * Server initialized
1:M 17 Nov 2024 11:53:47.612 * Ready to accept connections tcp
```

The hostname of the redis container is:
```
rockwurst:/ # docker exec -it broker /bin/bash
root@ecab3d971488:/data# hostname
ecab3d971488
```
so, not sure whether that is the issue, or what else I am missing.
Am I doing it wrong?

BrokenOnedroid:
@Hooverdan96 , 
I compared my settings against yours. 
Looks like i assigned the same network to all relevant containers.
Screenshot_20241118-083444

Expect the usage of
`--net=host`
I did not find any rockon which assigned a network on creation. I guess I assigned them all manually.
Can you check if attaching the containers to a docker network changes anything for you.

I can reach the webgui.
Screenshot_20241118-083553

Is it possible to create a network on installation of a Rockon?

FroggyFlox:
> Looks like i assigned the same network to all relevant containers.

I believe this is the needed part, indeed. The `redis://broker` bit relies on name resolution which doesn't exist on docker's default `bridge` network, but does on user-created one.

> Is it possible to create a network on installation of a Rockon?

Not yet, at least not directly, but this is something I'd like to see implemented as it becomes more and more necessary. This is especially true given @phillxnet's recent work on improving multi-containers rockons. A current alternative could be to use `container_links` to do just that. It's not as straightforward as setting a network manually at the moment, however, but itoght do the trick. If interested, I would look at @phillxnet's recent work on the BareOS rockon.

Hooverdan96:
I'll try this again with the network piece

Hooverdan96:
I was hesitatnt to recommend the container links, mostly for the fact that it's been considered legacy for some time. However, that certainly will make it easier for now.

I keep getting this error:
```
django.db.utils.OperationalError: connection failed: connection to server at "172.18.0.3", port 5432 failed: FATAL:  password authentication failed for user "paperless"
```
As you can see on the above, I've been using a different user/password combination in my config.

This implies to me that the postgres database/user/password needs to be what you've been using, i.e. `paperless` in all three entries.

Hooverdan96:
And when I look in the documentation, unless specified explicitly for the paperless container, it will all default to those values:

image

paperless-ngx.json
```diff
@@ -0,0 +1,113 @@
...
+                    "PAPERLESS_REDIS": {
+                        "description": "Set to redis://paperless-broker:6379.Redis connection URL. Do not change unless you know what you are doing.",
+                        "value": "redis://paperless-broker:6379",
+                        "label": "Redis URL"
+                    },
+                    "PAPERLESS_DBHOST": {
+                        "description": "Set to paperless-db. Hostname of the PostgreSQL database container.",
+                        "value": "paperless-db",
+                        "label": "Database Host"
+                    },
```
Hooverdan96:
```suggestion
```

paperless-ngx.json
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    "name": "paperless-to-paperless-db",
+                    "source_container": "paperless-db"
+                }
+            ]
+        },
+        "containers": {
+            "paperless": {
+                "image": "ghcr.io/paperless-ngx/paperless-ngx",
+                "launch_order": 3,
```
Hooverdan96, line 18:
```suggestion
                "launch_order": 3,
"opts": [
                    [
                        "-e",
                        "PAPERLESS_DBHOST=paperless-db"
                    ],
                    [
                        "-e",
                        "PAPERLESS_REDIS=redis://paperless-broker:6379"
                    ]
                ],
```
Since these values don't really need to change in this Rockon context, I suggest these can be moved into the options sections and so won't require a user to enter these values verbatim. Changes to these would likely require changes to the overall json anyway, and if an advanced user so desires, they will likely create their own version of this.

paperless-ngx.json
```diff
@@ -0,0 +1,113 @@
...
+                    "/data": {
+                        "description": "Choose a Share for Redis data.",
+                        "label": "Redis Data Share"
+                    }
+                }
+            },
+            "paperless-db": {
+                "image": "docker.io/library/postgres",
+                "tag": "16",
+                "launch_order": 2,
```
Hooverdan96, line 112:
```suggestion
                "launch_order": 2,
                "opts": [
                    [
                        "-e",
                        "POSTGRES_DB=paperless"
                    ],
                    [
                        "-e",
                        "POSTGRES_USER=paperless"
                    ],
                    [
                        "-e",
                        "POSTGRES_PASSWORD=paperless"
                    ]
                ],
```

Since these values seem to be the default for the paperless container itself, unless we also want to explicitly expose those values in the paperless container, I would suggest to go with the default values, and hence being able to hide them from user entry.

paperless-ngx.json
```diff
@@ -0,0 +1,113 @@
...
+                    "POSTGRES_USER": {
+                        "description": "Enter the username for the PostgreSQL database.",
+                        "label": "Database Username",
+                        "index": 2
+                    },
+                    "POSTGRES_PASSWORD": {
+                        "description": "Enter the password for the PostgreSQL database user.",
+                        "label": "Database Password",
+                        "index": 3
+                    }
```
Hooverdan96:
```suggestion
```

Move the environment variables into the `opts` section, since these don't need to be changed from the defaults.

paperless-ngx.json
```diff
@@ -0,0 +1,113 @@
...
+                }
+            }
+        },
+        "description": "Paperless DMS - Private & Secure document management system. <p>Based on a Docker image: <a href='https://github.com/paperless-ngx/paperless-ngx/tree/main'>GitHub Repository</a>. Available for amd64 and arm64 architectures.</p> <p> Create User like shown here (#7): <a href='https://docs.paperless-ngx.com/setup/#docker_hub'>https://docs.paperless-ngx.com/setup/#docker_hub</a> </p>",
```
Hooverdan96:
```suggestion
        "description": "Paperless DMS - A private & secure Document Management System. <p>Based on the official Docker image: <a href='https://github.com/paperless-ngx/paperless-ngx/tree/main'>GitHub Repository</a>, available for amd64 and arm64 architecture.</p> <p><i>Note:</i> Create Superuser like shown here (#7): <a href='https://docs.paperless-ngx.com/setup/#docker_hub'>https://docs.paperless-ngx.com/setup/#docker_hub</a> by using the command line and entering the paperless container with <code>docker exec -it paperless /bin/bash</code></p><p>Then execute <code>python3 manage.py createsuperuser</code>, which will then prompt for a username, email address and password.</p><p>These credentials can then be used to intially login into the paperless WebUI.</p>",
```
Making the superuser creation for initial login more explicit, in addition to the link you already mentioned.

Hooverdan96 (changes requested):
Thank you for your persistence in getting this to work. I have some further suggestions to further simplify the setup (mainly moving static environment values into the options sections of the respective containers).
I did not touch the time zone variable, since I am not sure whether the fact that Rockstor automatically forces any container timezone to the host timezone. But I suspect that this goes directly into the Django config, so for now, we can probably leave that one alone, unless @phillxnet or @FroggyFlox have an opinion on that.

Thanks again for working out the "trickiness" so that even someone like myself could finally install and run it!

Hooverdan96:
@BrokenOnedroid, I successfully installed it based on your latest push! Thanks for that.

Also, I proposed some suggestions (tested those on my own system) to move some of the variables into the options section to further reduce the requirement for the user to type out default values that should not be changed for this Rockon definition.

Take a look and let me know what you think.

Thanks again for this great submission!

Hooverdan96:
And apparently all of this I did before/during your last push where you spelled out the user/password info for the paperless container. Do you think it's important (aside from the superuser) to have the flexibility to change the user/password/redis pieces? Or is it more user-friendly to predetermine what these connections need to look like and not expose them?

phillxnet:
@Hooverdan96 Re:

> I was hesitant to recommend the container links, mostly for the fact that it's been considered legacy for some time.

Pretty sure @FroggyFlox moved this older Rock-on links definition mechanism to instead use/setup docker network some time ago:

See: "Implement docker networks. Fixes #1982." #2207

Confusing and potentially obfuscating to re-use the links syntax but it played well for minimising definition changes and the associated supporting code. At least that was the thinking at the time.  

> As detailed in #1982, our current inter-containers communication in multi-containers rock-ons is not functional due to the deprecation of containers links by Docker. As originally proposed by @flukejones, we should switch to using docker networks instead.
> This pull-request thus represents a fix for inter-containers communication by following @flukejones idea and recommendation. It also takes the opportunity to expand upon this and proposes a deeper integration of Docker networks into Rockstor.

Looks to have been rolled out, rpm wise, since: https://github.com/rockstor/rockstor-core/releases/tag/4.0.5-0

So it is also available in our our last two Stable releases 4.1.* & 4.6.*. My apologies for not making a bigger deal of this. It was one of the major feature in 4.1.* and was a major Rock-on improvement (saving the whole subsystem re multi-container arrangements) . Plus it got us out from a hard deprecation. My efforts (recently stalled but to be resumed shortly) in the BareOS rock-on were partly intended to show-case this capability a little more.

See: 

Take a look at @FroggyFlox excellent PR for this as there is loads of explanation there and proof of function etc. As always, and as per your recent issue, we can always improve things; but this was a major improvement re multi-container Rock-ons. We just need a similar effort to support the newer docker storage capability (volumes)

See: https://docs.docker.com/engine/storage/volumes/

However; all in good time :).

BrokenOnedroid:
@Hooverdan96 I'm going to take a look at your work, but I don't have the time right now.
Personally I would prefer to be able to set Username and Password. A fixed Redis is better. 
But ofc for usability, less work is better.

Hooverdan96:
@BrokenOnedroid thanks. I probably took it to the extreme on trying to get rid of definable items where possible. So, if you think being able to set the user/password combination (and that's currently unfortunately then required twice, once for paperless and then for the postgresql container) provides additional value for other reasons, then keep them out. I didn't think the django to database connection in paperless would be that relevant for exposing (thinking about Rockstor where that connection is pre-defined as well.
Defining the superuser during setup (as you now added) is definitely more convenient, but also less secure, since it's stored in the database (and visible in cleartext) compared to going into the container and create it via the command line. However, for a more casual user, this requires no interaction with the command line, so I think that's better in the end.

But ... when you have time again, let us know what you think. If you do decide that the db user/pwd should be exposed, I'd say make it clear in the description that they need to match. On the redis link we seem to agree that it in this context it makes sense to move it to the options, so the user doesn't have to populate it.

@phillxnet thanks for summarizing the links topic nicely! I do remember this now (but apparently didn't make the connection).

BrokenOnedroid:
@Hooverdan96 i combined our changes. Setting the focus on usability.

Hooverdan96:
@BrokenOnedroid thanks for the merged changes! I am trying to test right now, but getting some timeout on one of the layers when it tries to pull the redis image (and hence the overall install fails). I will try again later, hopefully that upstream issue will be resolved by then

Hooverdan96:
ok, I was able to run through the installation procedure 
image

and after some wait time, the installation completed successfully. I was able to reach the UI screen and successfully log in.

image

I did not execute any of the paperless functionality at this time.

@phillxnet, @FroggyFlox from my perspective this is working and can accepted and merged.

phillxnet:
@Hooverdan96 Re:

> @phillxnet, @FroggyFlox from my perspective this is working and can accepted and merged.

I can merge and publish, but GitHub currently shows an unresolved change request made in your review. Can you do the 3-dots next to the "1 change requested" at the bottom on this PR to ensure we are on the same page here. Let me know if you need more rights to do this: but pretty sure you have all that are required here given your Rock-ons repo maintainer status. From a quick look @BrokenOnedroid has addressed all that has been brought up here. But I just want to be sure before merging & publishing.

@BrokenOnedroid Thanks for persevering with this Rock-on submission. Much appreciated.

Hooverdan96 (approved):

phillxnet:
As per @Hooverdan96 review I too have a successful install and Web-UI access; with a little delay for me before the Web-UI was available.

Initial-install-paperless-ngx

As such I'll merger this and get it published very shortly.

---

I have some recommendations re our Rock-on share labels within the Web-UI but I'll leave them until we have a proof of concept re the pending BareOS rock-on work. Briefly I think we should have suggested share names included. And remove redundant text such as we have in this and many other existing Rock-ons.

But lets get this merged and out as-is for now as it appears to work, which supersedes all cosmetics.

Thanks again @BrokenOnedroid and @Hooverdan96 for making this Rock-on a reality.

phillxnet:

Hooverdan96:
yes, the WebUI delay seems normal, I've experienced that as well. I assume it takes some time for all the components to be aligned and all that. And you're probably correct on the labels/descriptions.

PR #398, opened 2024-12-07
Linkwarden

BrokenOnedroid:
- name: Linkwarden
- website: https://linkwarden.app/
- description: Linkwarden is an open-source collaborative bookmark manager to collect, organize and preserve webpages.

- docker image: https://github.com/linkwarden/linkwarden/pkgs/container/linkwarden)

### Checklist

### DIFF
--- a/linkwarden.json
+++ b/linkwarden.json
@@ -1,0 +1,70 @@
+{
+    "Linkwarden": {
+        "description": "<p>Linkwarden is a self-hosted, open-source collaborative bookmark manager for organizing and archiving webpages.</p><p>Based on the official image from <a href='https://github.com/linkwarden/linkwarden' target='_blank'>Linkwarden</a>, compatible with amd64 and arm64 architectures.</p>",
+        "version": "0.9",
+        "website": "https://linkwarden.app/",
+        "containers": {
+            "linkwarden": {
+                "image": "ghcr.io/linkwarden/linkwarden",
+                "tag": "latest",
+                "launch_order": 2,
+                "ports": {
+                    "3000": {
+                        "description": "The port for accessing the Linkwarden Web UI.",
+                        "label": "Web UI Port",
+                        "host_default": 3000,
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/data/data": {
+                        "description": "Choose a Share for Linkwarden data storage.",
+                        "label": "Linkwarden Data Storage"
+                    }
+                },
+                "environment": {
+                    "DATABASE_URL": {
+                        "description": "PostgreSQL connection string. Set to postgresql://postgres:<password>@linkwarden-postgres:5432/linkwarden",
+                        "label": "PostgreSQL URL",
+                        "index": 3
+                    },
+                    "NEXTAUTH_SECRET": {
+                        "description": "A secure secret phrase.",
+                        "label": "NextAuth Secret",
+                        "index": 2
+                    },
+                    "NEXTAUTH_URL": {
+                        "description": "Base URL for Linkwarden. Set to: http://localhost:3000/api/v1/auth",
+                        "label": "NextAuth URL.",
+                        "index": 1
+                    }
+                }
+            },
+            "linkwarden-postgres": {
+                "image": "postgres",
+                "tag": "17-alpine",
+                "launch_order": 1,
+                "volumes": {
+                    "/var/lib/postgresql/data": {
+                        "description": "Choose a Share for PostgreSQL database storage.",
+                        "label": "PostgreSQL Storage"
+                    }
+                },
+                "environment": {
+                    "POSTGRES_PASSWORD": {
+                        "description": "Enter the password for the PostgreSQL 'postgres' super-user.",
+                        "label": "PostgreSQL Password"
+                    }
+                }
+            }
+        },
+        "container_links": {
+            "linkwarden": [
+                {
+                    "name": "linkwarden-to-linkwarden-postgres",
+                    "source_container": "linkwarden-postgres"
+                }
+            ]
+        }
+    }
+}
--- a/root.json
+++ b/root.json
@@ -32,6 +32,7 @@
     "JenkinsCI": "jenkins.json",
     "Kanboard": "kanboard.json",
     "LazyLibrarian": "lazylibrarian.json",
+    "Linkwarden": "linkwarden.json",
     "Logitech Media Server - Latest Stable": "lms_latest.json",
     "MariaDB": "mariadb.json",
     "Medusa": "medusa.json",


Hooverdan96:
I have tested the installation:
image
image
image
image
image

After the install completed, I needed to create a new account, and subsequently was able to log in:
image

So, this works!

Since there are field-interdependencies w.r.t. password and port setting, I am suggesting some additional comments in the field descriptions. If you agree you should be able to just accept the suggestions directly from this thread.

linkwarden.json
```diff
@@ -0,0 +1,70 @@
...
+                        "label": "PostgreSQL URL",
+                        "index": 3
+                    },
+                    "NEXTAUTH_SECRET": {
+                        "description": "A secure secret phrase.",
+                        "label": "NextAuth Secret",
+                        "index": 2
+                    },
+                    "NEXTAUTH_URL": {
+                        "description": "Base URL for Linkwarden. Set to: http://localhost:3000/api/v1/auth",
```
Hooverdan96, line 37:
```suggestion
                        "description": "Base URL for Linkwarden. Set to: http://localhost:<Web UI port defined earlier>/api/v1/auth",
```

linkwarden.json
```diff
@@ -0,0 +1,70 @@
...
+                    "3000": {
+                        "description": "The port for accessing the Linkwarden Web UI.",
+                        "label": "Web UI Port",
+                        "host_default": 3000,
+                        "ui": true
+                    }
+                },
+                "volumes": {
+                    "/data/data": {
+                        "description": "Choose a Share for Linkwarden data storage.",
```
Hooverdan96, line 21:
```suggestion
                        "description": "Choose a Share for Linkwarden data storage. E.g., create a share named linkwarden-data.",
```

linkwarden.json
```diff
@@ -0,0 +1,70 @@
...
+                "launch_order": 1,
+                "volumes": {
+                    "/var/lib/postgresql/data": {
+                        "description": "Choose a Share for PostgreSQL database storage.",
+                        "label": "PostgreSQL Storage"
+                    }
+                },
+                "environment": {
+                    "POSTGRES_PASSWORD": {
+                        "description": "Enter the password for the PostgreSQL 'postgres' super-user.",
```
Hooverdan96, line 55:
```suggestion
                        "description": "Enter the password for the PostgreSQL 'postgres' super-user that was entered previously as part of the 'PostgreSQL URL'",
```

Hooverdan96 (commented):
I assume, we really don't have many options to prevent the duplicate entry/dependencies between the WebUI port and the BaseURL, nor for the Postgres superuser password and the connection string. Only for the Postgres connection I could see it to move to the `opts` object, unless you think there's a good use case for knowing the super user password for the database.

Hooverdan96:
@BrokenOnedroid if you can review and commit the suggested changes, then I think this will be ready for final review and can be merged.

BrokenOnedroid:
@Hooverdan96 done

Hooverdan96 (approved):
This looks good, the tooltips provide the right clarity I think. Another test installation seemed to have been successful (none of the previous configuration items changed, only text elements.

Hooverdan96:
@phillxnet when you get a chance to merge and publish this. Thanks again @BrokenOnedroid for submitting this.

phillxnet:
@BrokenOnedroid & @Hooverdan96 Thanks for preparing this Rock-on. It looks like a nice addition.

It does rather highlight some of our current limitations however: with the install being super fragile re by-hand copying from tooltip suggestions. But this can all be improved upon as we up our Rock-on facilities over time. And with some pending examples of a preferred presentation that we hope to form into a guide regarding label naming and a new format to be expected across all Rock-on labels.

The `DATABASE_URL` was particularly challenging re input here. Especially with its associated password appearing later in the list (again a limitation on our end). It was also not clear what NEXTAUTH_* was all about. I.e. we have not user level explanation here and that is what Rock-ons are about in essence. I.e. "A secure secret phrase." for what?

I.e. on the label front:
- NEXTAUTH_SECRET: "NextAuth Secret"
- NEXTAUTH_URL: "NextAuth URL"

I know one can look these up - but in part our place as a wrapper is not to have the env vars presented bare with nothing but an upstream reference to look them up. We try to explain what each is for. If it's required to surface the env var we are setting then so be it: i.e. it's technical and so one needs to inform the user what element in upstream it references. But we do present a summary (pre-install) of how we translate what is input to the docker level.

Again function trumps form, and this looks to work. Plus it's far easier to refine something that exists than to have to start from scratch. So we are at least started here and I'll merge and publish accordingly. But more Rock-ons with self inconsistent labels/suggestions etc. And only a case shift of an environmental variable as presented at the docker command level rather makes this more tricky than a copy-past of a docker command. Ideally we need to have a little more refinement.

But all in, we have a start. But it represents, not like a goodly number of our other Rock-ons mind :), a potential technical dept. I.e. there is work to be done on the user-text side. But we also have work to do in communicating our preferred format for example labels. And as always this is ongoing: we we are getting there.

For all involved: do not take this overly negatively, we are just moving to up our presentation game also here. And for that we need a doc guide on what is to be expected for labels etc.

However - nice to have this Rock-on submitted and in place (about to merge and publish). And thanks again for all efforts getting it this far.

phillxnet:
@Hooverdan96 We could likely use some kind of random password/passphrase place-holder: which when encountered in a Rock-on definition is substituted for an actual random pass. And can be re-used multiple times in a Rock-on definition. Probably needs a new named element with some definition on the type of password/passphrase to generate. Tricky to be backward compatible here thought. But we could introduce some minimum Rockstor version element: surfaced to the user. Maybe something for the next or next but one testing channel phase.

phillxnet:
PRODUCTION published

Hooverdan96:
@phillxnet agreed on your points. That has always been a challenge, that one cannot really copy from the tooltips. I think it would make sense, at the least, to introduce defaults for all elements that can be surfaced (I think we have an issue open related to this). That way the copy/paste could be avoided or at least minimized. Even if the default shows (like in this one) some element that has to be changed for sure, because of dependency.

Secondly, based on your thoughts above, maybe introduce another json element, as optional, that can be handled like a variable, i.e. it's exposed once, the user defines it, and then can be mapped into other containers and their variables. Making that optional would possibly allow the backwards compatibility and give time to eventually convert all Rockons over time to take advantage of these things where applicable.

Hooverdan96:
With regards to default tags, the rockstor-core repot, we currently have:
[#2094](https://github.com/rockstor/rockstor-core/issues/2094)
[#2012](https://github.com/rockstor/rockstor-core/issues/2012)

phillxnet:
@Hooverdan96 Re:

> ... maybe introduce another json element, as optional, that can be handled like a variable, i.e. it's exposed once, the user defines it, and then can be mapped into other containers and their variables.

Well put. Yes; that would be super handy, and along with a default=autopassword, or the like and it can pre-set an internal password between containers such as is more common re all these postgres containers etc. We could also have it internal only (optionally) so there is no interaction/input required if need be.

PR #405, opened 2024-12-09
Update paperless-ngx.json

BrokenOnedroid:
I made a mistake in the volume part for paperless. I have fixed this. Should have no impact on usage and update as only the export and import shares were changed. Added PAPERLESS_SECRET_KEY, PAPERLESS_URL and PAPERLESS_CSRF_TRUSTED_ORIGINS which can all be left empty but are useful if a nginx or similar is in use.

### DIFF
--- a/paperless-ngx.json
+++ b/paperless-ngx.json
@@ -36,19 +36,19 @@
                     }
                 },
                 "volumes": {
-                    "/data": {
+                    "/usr/src/paperless/data": {
                         "description": "Choose a Share for data file storage.",
                         "label": "Config paperless Data - Storage"
                     },
-                    "/media": {
+                    "/usr/src/paperless/media": {
                         "description": "Choose a Share for media storage.",
                         "label": "Config paperless Media - Storage"
                     },
-                    "/export": {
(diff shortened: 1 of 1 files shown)


Hooverdan96:
Glad you found the volume issue, because our previous tests didn't really execute the functionality but only the installation success. So that change should be fine to go.

On the additional fields though, we still have the same challenge that the environment fields are currently all mandatory with no way of setting them to optional. So, if at this time, we were to add these fields in, a user would have to maintain all of that info even if no public-facing implementation is planned. Related to this [#2280](https://github.com/rockstor/rockstor-core/issues/2280).

image

So for now, I recommend to split this PR, park the additional environment fields, and just push the immediate changes required. We can then revisit (likely after the next stable release) to see how this can be addressed. Of course, you're welcome to take a look yourself.

BrokenOnedroid:
I removed the env vars. So the volumes can be fixed.

Hooverdan96 (approved):
Looks good to me. re-installation showed the mapped volumes working properly.

phillxnet:
PRODUCTION published

PR #415, opened 2024-12-16
Deprecate Zeronet (#414)

### LINKED ISSUE #414: Deprecate Zeronet
Both custom docker image as well as the official repository has not seen any updates in 3-4 years, unfortunately.

Hence suggesting to deprecate this Rockon

Hooverdan96:
no updates for 3 years to custom docker image as well as repository

Fixes #414  .

### Checklist
- [X] Entry removed from `root.json`


phillxnet (approved):
@Hooverdan96 Thanks again for taking care of yet-another Rock-on removal. Much appreciated.

And well done remembering that comma removal also :) .

phillxnet:
PRODUCTION published

phillxnet:
@Hooverdan96 This and the other Rock-on removal has the test system I have here to below 6 seconds for Rock-on info retrieval:

```
[17/Dec/2024 19:17:10] INFO [storageadmin.views.rockon:504] Rock-on definitions retrieved in: 5.86 seconds.
```

It's been creeping up of late, re all these new additions. So always good to do some clear-out to balance this. We can't support an endless number of official Rock-ons.

Hooverdan96:
Maybe as part of future dev, we can consider more of a delta pull or something (adding a "last downloaded"/"last updated" timestamp or something).

phillxnet:
@Hooverdan96 Maybe, but we have to avoid complexity at almost all costs is my thinking. That last change we did re holding the ssl info really helped. Maybe just switching to a faster web-get method would do us. Plus I think there is server-side stuff we may be able to enable now also. But I'm concerned it will fail with older requests. That may also be a better route as we then server-side zip and cache. All pros and cons. We just need to keep an eye on that figure as we add more and more Rock-ons. But deleting them certainly helps :) .

Issue #422, opened 2025-01-30
Integrate Rockon validator to check Rockon related PRs


Hooverdan96:
Since @StephenBrown2 donated his work on Rockon json file validation (https://github.com/rockstor/rockon-validator), there could be an opportunity as suggested by @phillxnet to include some high-level validation via Github actions when new PRs are submitted (or updates to existing PRs are pushed). This in turn could then lead to additional checks for the rockon-validator to make submissions more streamlined.

PR #423, opened 2025-01-30
Create Rockon validator action (#422)

Hooverdan96:
Addresses #422 .

This pull request proposes to add a GitHub action to essentially provide some linting and formal review of newly submitted Rockons or subsequent pushes to update an existing PR. It will only apply to json files.


Hooverdan96:
@phillxnet, @FroggyFlox , I have done some testing in a dummy repository on this (I'll post some of the results later). Do you think this is a good idea?

.github/workflows/rockon_validator.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+# show differentials based on ideal Rockon schema and whether it's in the root.json file
+      - name: Process all changed json files
+        if: steps.changed-json-files.outputs.any_changed == 'true'
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-json-files.outputs.all_changed_files }}
+        run: |
+          for jsonfile in ${ALL_CHANGED_FILES}; do
+            rockon-validator -d $jsonfile
+          done
```
StephenBrown2, line 57: If this is just meant to be informative, and not fail if the spec isn't met, this is fine. If you want to fail the check if there is a diff to be changed, you can add the `-c, --check` flag.

Additionally, the tool can parse multiple files at once, so the step can be simplified to the following:

```suggestion
        run: rockon-validator -c -d ${{ steps.changed-json-files.outputs.all_changed_files }}
```

Hooverdan96: @StephenBrown2 I experimented with the `-c` flag, but figured that to start with it would be easier to look at the output as a guidance to what would look different. thanks for the suggestion, I just couldn't make it work originally with multiple files (as I was doing it wrong to begin with ... ) hence I arrived at the above.
qq since I have your attention already: can I also add the root file check into the same command line?

phillxnet:
> Do you think this is a good idea?

Definitely. It would be great to finally get @StephenBrown2 prior efforts here into a PR GitHub action. Although both you and @FroggyFlox are more informed on the entire process, so I can only wave enthusiastically currently.

.github/workflows/rockon_validator.yaml
StephenBrown2: Sorry my attention is not very good these days. ;-)

The root file check is always done, passing `--root <path>` will just tell the tool to use a different `root.json` file. Let me know if there's anything I should change in https://github.com/rockstor/rockon-validator/tree/main?tab=readme-ov-file#rootjson to make the usage clearer.

Hooverdan96: no worries at all. Thanks for the clarification. I wasn't entirely clear on that. I will post another question in a separate thread to you on the docker option that you provided.

phillxnet:
@Hooverdan96 Am I the hold on on this being ready to merge? If I understant this correctly - it is advisory only. And we need it in-play to know if it is advising us usefully.

Hooverdan96:
@phillxnet, no, you're not the holdup. I have not spent recent quality time with it. And you are correct, it would be advisory, at the same time we might want to reduce a check or two (mostly on the sequence of the tags) - and I have not figured out yet how to make this an informational vs. an error, if that's what we want.
I hope, I'll get more time to play with the github actions + the go repo

phillxnet:
@Hooverdan96 Re:

> at the same time we might want to reduce a check or two (mostly on the sequence of the tags)

Having now gotten a little more familiar with our rockon-validator, the above is actually quite tricky - doable in time but we may just have to take what we can get for now re validation with the downside of strange ordering re specifically your:

- https://github.com/rockstor/rockon-validator/issues/13

The reason is that the original json is ingested and normalised against a GO structure which is then exported back out to json to form the suggested/normalised 'format'.

My thinking here was to use the --diff output of the rockon-validator to create a patch, which is then presented as a suggested change to the associated Pull request. There-by addressing all the format changes that the rockon-validator can handle to leave us with less to address there-after.

PR #428, opened 2025-03-23
Update pi-hole to version 6.x (#427)

### LINKED ISSUE #427: Pihole v6 update
Hey there, I think we may need to update the pihole rockon in order to be compatible with the latest and greatest pihole v6.

https://pi-hole.net/blog/2025/02/18/introducing-pi-hole-v6/
_You can start the container with the old variables in place but don’t expect them to work! It is recommended to read the docker section of our docs page before upgrading._

https://docs.pi-hole.net/docker/

I haven't attempted to upgrade pihole on my live server yet.

Hooverdan96:
Fixes #427  .
 - Address breaking changes between pi-hole v5.x and v6.x
 - some text cleanup, convert to Unix LR and replace tabs with spaces.

- docker image: https://hub.docker.com/r/pihole/pihole

### Checklist

### DIFF
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -14,7 +14,7 @@
                     ],
                     [
                         "-e",
-                        "FTLCONF_misc_etc_dnsmasq_d: 'true'"
+                        "FTLCONF_misc_etc_dnsmasq_d=true"
                     ]
                 ],
                 "launch_order": 1,
--- a/pi-hole.json
+++ b/pi-hole.json
@@ -1,7 +1,7 @@
 {
-    "Pi-Hole": {
+    "Pi-Holev6": {
         "containers": {
-            "pi-hole": {
+            "pi-holev6": {
                 "image": "pihole/pihole",
                 "opts": [
                     [
@@ -15,6 +15,10 @@
                     [
                         "-e",
                         "FTLCONF_misc_etc_dnsmasq_d=true"
+                    ],
+                    [
+                        "-e",
+                        "DNSMASQ_USER=root"
                     ]
                 ],
                 "launch_order": 1,
@@ -32,7 +36,7 @@
                         "label": "Pi-Hole config"
                     },
                     "/etc/dnsmasq.d": {
-                        "description": "Choose a share for dnsmasq configuration. E.g.: create a Share called dnsmasq-config for this purpose alone.",
+                        "description": "Choose a share for dnsmasq configuration. E.g.: create a Share called dnsmasq-config for this purpose alone. The share needs to have full access for a user with a corresponding UID=1000",
                         "label": "dnsmasq config"
                     }
                 },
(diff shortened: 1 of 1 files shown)


Hooverdan96:
Here's my cursory testing of the new Rockon version for pi-hole version 6:

(temporary title change to avoid conflict with standard Rockon list):
image

image

Rockon Web Port (i.e. allowing to access Rockon page directly from within Rockstor - see tool tip):
image

actual pi-hole web-port (needs to be the same) and password - also see tool tip
image

Installation summary:
image

Installation successful:
image

Accessing login page from within Rockstor:

image


Access after putting login password:
image

Default DNS upstream servers are active, and settings can be adjusted from within WebUI.

pi-hole.json
```diff
@@ -1,75 +1,62 @@
...
+                        "--cap-add",
+                        "NET_ADMIN"
+                    ],
+                    [
+                        "--net",
+                        "host"
+                    ],
+                    [
+                        "-e",
+                        "FTLCONF_misc_etc_dnsmasq_d: 'true'"
```
kanecko, line 17: I haven't actually tested the rockon, but according to refdoc, the -e expects the following format: "VAR=VAL".
https://docs.docker.com/reference/cli/docker/container/run/#env

pi-hole.json
```diff
@@ -1,75 +1,62 @@
...
-						"--dns",
-						"8.8.8.8"
-					],
-					[
-						"--net",
-						"host"
-					],
-					[
-						"-e",
-						"IPv6=False"
```
kanecko, line 25: does the omission of this VAR means that IPv6 functionalities will be enabled?

pi-hole.json
```diff
@@ -1,75 +1,62 @@
...
+                    "/etc/pihole": {
+                        "description": "Choose a share for Pi-Hole configuration. E.g.: create a Share called pihole-config for this purpose alone.",
+                        "label": "Pi-Hole config"
+                    },
+                    "/etc/dnsmasq.d": {
+                        "description": "Choose a share for dnsmasq configuration. E.g.: create a Share called dnsmasq-config for this purpose alone.",
+                        "label": "dnsmasq config"
+                    }
+                },
+                "environment": {
```
kanecko, line 39: Just a question since I don't know the inner-workings of Rockstor.
Was the following env option and comment from the default composer file considered?
      # If using Docker's default `bridge` network setting the dns listening mode should be set to 'all'
      FTLCONF_dns_listeningMode: 'all'

pi-hole.json
Hooverdan96: good point. I didn't get an error message and followed the instruction listed here:

https://docs.pi-hole.net/docker/upgrading/v5-v6/

however, I just realized that this is probably for when one uses docker-compose and not just the docker run command. I'll fix it.

pi-hole.json
Hooverdan96: Since the Rockon is set to use the host networking stack (set with environment variable `net=host` this option doesn't apply. The host networking is required so that if one uses pi-hole also as a dhcp server (not just a dns sink) they can.

pi-hole.json
```diff
@@ -1,75 +1,76 @@
...
+                        "-e",
+                        "FTLCONF_misc_etc_dnsmasq_d=true"
+                    ],
+                    [
+                        "-e",
+                        "DNSMASQ_USER=root"
+                    ]
+                ],
+                "launch_order": 1,
+                "ports": {
```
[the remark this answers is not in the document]
Hooverdan96: since this version is using `net=host` you don't need to really specify any ports for this. It will be visible:

https://docs.pi-hole.net/docker/DHCP/#docker-pi-hole-with-host-networking-mode

phillxnet (approved):
@Hooverdan96 Thanks for seeing to this important update. My apologies for the delay here.
@kanecko Thank for your initial review. Much appreciated.

confirm-install-settings


I'm just doing a quick review give @Hooverdan96 extensive testing and that we have already had a review here from @kanecko 

Installed OK with the already indicated :
```
Mar 28 11:14:50 rleap15-6 pihole/pihole:latest/pi-hole[20644]: 2025-03-28 11:14:46.501 WET [181M] WARNING: Insufficient permissions to set process priority to -10 (CAP_SYS_NICE required), process priority remains at 0
Mar 28 11:14:50 rleap15-6 pihole/pihole:latest/pi-hole[20644]: 2025-03-28 11:14:46.503 WET [181M] WARNING: Starting pihole-FTL as user root is not recommended
...
Mar 28 11:14:50 rleap15-6 pihole/pihole:latest/pi-hole[20644]: 2025-03-28 11:14:47.245 WET [181M] WARNING: Insufficient permissions to set system time (CAP_SYS_TIME required), NTP client not available
Mar 28 11:14:50 rleap15-6 pihole/pihole:latest/pi-hole[20644]: 2025-03-28 11:14:47.246 WET [181M] INFO: FTL is running as root
```
From the tooltip/help text on the associated shares - it suggested only one share need be accessible by the UID/GID. I'm unsure if this was intentional. In my test install I set both shares similarly.

Web-UI button works and we have the following versions reported.

Docker Tag [2025.03.0](https://github.com/pi-hole/docker-pi-hole/releases/2025.03.0)Core [v6.0.5](https://github.com/pi-hole/pi-hole/releases/v6.0.5)FTL [v6.0.4](https://github.com/pi-hole/FTL/releases/v6.0.4)Web interface [v6.0.2](https://github.com/pi-hole/web/releases/v6.0.2)

At this point my /mnt2/dnsmask-config/ has no content.
But /mnt2/pihole-config does, with all top-level file/dirs bar `cli_pw`, `hosts` & `logrotate`,  `pihole-*`, & `versions` owned by the selected test:users. Indicated files were root:root.

I'll get this merged directly and published soon there after.

Hooverdan96:
> From the tooltip/help text on the associated shares - it suggested only one share need be accessible by the UID/GID. I'm unsure if this was intentional. In my test install I set both shares similarly.

Yes, the PUID/GID only "works" on pihole itself. The user for the dnsmasq functionality is a root user still. I've tried enabling the change for that user as well, but that didn't work, unless one changes the user to `root`. So for dnsmasq the share can be whatever one assigns.

<At this point my /mnt2/dnsmask-config/ has no content.
This is related to that if someone wants to place a specific dnsmasq config in there, it will read from there. Otherwise it's all contained within the pi-hole container/db. So, more of a legacy thing it seems, to be able to use some more of the exotic settings for a dhcp server.

phillxnet:
PRODUCTION published

PR #434, opened 2025-03-30
Support alternate web UI for Transmission OpenVPN

### LINKED ISSUE #433: Update Transmission with OpenVPN to support alternate web UIs
The Transmission with OpenVPN image includes support for alternate web UIs for Transmission through the use of an environment variable. [See docs here](https://haugene.github.io/docker-transmission-openvpn/config-options/#alternative_web_uis)

It would be helpful to expose the `TRANSMISSION_WEB_UI` setting in Rockstor to allow choosing one of these alternates.

jsutlovic:
Adds the `TRANSMISSION_WEB_UI` environment variable to config for the `Transmission - OpenVPN` image with documentation based on the [official image docs](https://haugene.github.io/docker-transmission-openvpn/config-options/#alternative_web_uis). 

Tried this with both `default` as the setting, as well as `combustion`, `flood-for-transmission` and `transmissionic` and it works fine. Users that re-install the image will just need to add `default` for the variable to continue with the existing experience.

I've also updated the rock-on version and website to link to the project website rather than the docker image itself.

Fixes #433 .

This pull request updates the existing `transmission-with-openvpn` rock-on. 

### Checklist

### DIFF
--- a/transmission-with-openvpn.json
+++ b/transmission-with-openvpn.json
@@ -72,15 +72,20 @@
                         "label": "LOCAL_NETWORK",
                         "index": 6
                     },
+                    "TRANSMISSION_WEB_UI": {
+                        "description": "Alternative web UI to use; Use 'default' to leave the default Transmission UI, other options are: combustion, kettu, transmission-web-control, flood-for-transmission, shift or transmissionic.",
+                        "label": "TRANSMISSION_WEB_UI",
+                        "index": 7
+                    },
                     "PUID": {
                         "description": "Choose a User ID to run Transmission as",
                         "label": "User ID",
-                        "index": 7
+                        "index": 8
                     },
                     "PGID": {
                         "description": "Choose a Group ID to run Transmission as",
                         "label": "Group ID",
-                        "index": 8
+                        "index": 9
                     }
                 }
             }
--- a/transmission-with-openvpn.json
+++ b/transmission-with-openvpn.json
@@ -1,7 +1,7 @@
 {
     "Transmission - OpenVPN": {
-        "website": "https://hub.docker.com/r/haugene/transmission-openvpn/",
-        "version": "1.1",
+        "website": "https://haugene.github.io/docker-transmission-openvpn/",
+        "version": "1.2",
         "description": "Docker container running Transmission torrent client with WebUI while connecting to OpenVPN.<p>Based on a custom docker image: <a href='https://hub.docker.com/r/haugene/transmission-openvpn' target='_blank'>https://hub.docker.com/r/haugene/transmission-openvpn</a>, available for amd64 and arm64 architecture.</p>",
         "more_info": "<p>See the container's documentation for a list of included VPN provider profiles. If your provider isn't included, set it up as a custom provider, by putting the relevant files in the <code>Custom profile</code> share.</p>",
         "containers": {
--- a/transmission-with-openvpn.json
+++ b/transmission-with-openvpn.json
@@ -25,11 +25,11 @@
                 "volumes": {
                     "/data": {
                         "description": "Choose a Share where Transmission will save all of it's config and data files including your downloads. Eg: create a Share called transmission-data.",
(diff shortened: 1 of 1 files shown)


transmission-with-openvpn.json
```diff
@@ -72,15 +72,20 @@
                         "label": "LOCAL_NETWORK",
                         "index": 6
                     },
+                    "TRANSMISSION_WEB_UI": {
+                        "description": "Alternative web UI to use; Use 'default' to leave the default Transmission UI, other options are: combustion, kettu, transmission-web-control, flood-for-transmission, shift or transmissionic.",
```
Hooverdan96, line 76:
```suggestion
                        "description": "Alternative web UI to use; Use 'default' to leave the default Transmission UI, other options are (enter in lowercase): combustion, kettu, transmission-web-control, flood-for-transmission, shift or transmissionic.",
```
Since there is an emphasis on uppercase in the Provider variable, it might make sense to indicate that these should be lowercase (I did not test whether it matters, but since the values are listed like that in the documentation).

Hooverdan96 (changes requested):
Not your fault, but could you adjust this section:

```
                    "/data": {
                        "description": "Choose a Share where Transmission will save all of it's config and data files including your downloads. Eg: create a Share called transmission-data.",
                        "label": "Data Storage"
                    },
```
with a couple of corrections to this:
```
                    "/data": {
                        "description": "Choose a Share where Transmission will save all of its config and data files including your downloads. E.g.: create a Share called transmission-data.",
                        "label": "Data Storage [e.g. transmission-data]"
                    },
```
and here:
from:
```
                    "/etc/openvpn/custom": {
                        "description": "Choose a Share where a set of custom OpenVPN profile files can be found. This can be an empty share if you use one of the included provider profiles.",
                        "label": "OpenVPN config"
                    }
```
to:
```
                    "/etc/openvpn/custom": {
                        "description": "Choose a Share where a set of custom OpenVPN profile files can be found. This can be an empty share if you use one of the included provider profiles.",
                        "label": "OpenVPN config [e.g.: ovpn-config]"
                    }
```

transmission-with-openvpn.json
jsutlovic: I believe the lowercase does matter, so this is a good point. I'll add the modification :+1:

Hooverdan96: For this portion you can just select the "Commit Suggestion" button up there, and it will automatically merge/push it as an update (you'll see it on your side and in this PR). For the other suggestions I had, I couldn't figure out how to do that.

jsutlovic:
@Hooverdan96 Thanks for the quick review! Wasn't able to get to these updates over the weekend, but I've made the changes you requested and added an extra detail to clarify that `LOCAL_NETWORKS` supports a comma-separated list (official docs mention it [here](https://haugene.github.io/docker-transmission-openvpn/config-options/#network_configuration_options), and I have used it successfully myself).

Hooverdan96 (approved):
@jsutlovic thanks for the updates!
I did another install test with them and also added two networks as `LOCAL_NETWORK`.

image

After installation, checking the logs, both entries are picked up and routes are added:
```
adding route to local network 192.168.1.0/24 via 172.17.0.1 dev eth0
adding route to local network 172.16.0.0/24 via 172.17.0.1 dev eth0
```
So, from my perspective, this works out. I did not test an actual transmission download, but I am assuming at this time, that it will work (since you've been using it alread).

Well done, and thank you again for helping keeping the Rockons up to date!

phillxnet:
@Hooverdan96 & @jsutlovic My apologies for dropping the ball on this one, regarding the unreasonable delay in doing the final publishing bit.

@Hooverdan96 Thanks for looking to the review here, and getting the merge done in a timely manner; unlike myself on the final publishing step! I've tested a systems ability to read the entire repo as it stand post merge and no errors were reported, and the new changes were evident in the affected Rock-on. Moving to publish what we now have here to our production server.

PR #442, opened 2025-05-10
Restic rest-server

ilium007:

Fixes # .

- name: Rest rest-server
- website: https://github.com/restic/rest-server
- description: Rest Server is a high performance HTTP server that implements restic's [REST backend API](https://restic.readthedocs.io/en/latest/100_references.html#rest-backend). It provides secure and efficient way to backup data remotely, using [restic](https://github.com/restic/restic) backup client via the [rest: URL](https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#rest-server). This new rock-on requires a share to be specified to hold the Restic backup repository.

- docker image: [<link to image page on docker hub>](https://hub.docker.com/r/restic/rest-server)

### Checklist

### DIFF
--- a/rest-server.json
+++ b/rest-server.json
@@ -1,0 +1,50 @@
+{
+    "rest-server": {
+        "containers": {
+            "rest-server": {
+                "image": "restic/rest-server",
+                "tag": "latest",
+                "launch_order": 99,
+                "ports": {
+                    "8000": {
+                        "description": "Incoming REST server port. The Restic rest-server default (TCP 8000) is already taken so TCP 9000 is used here.",
+                        "host_default": 9000,
+                        "label": "REST server port",
+                        "protocol": "tcp"
+                    }
+                },
+                "volumes": {
+                    "/data": {
+                        "description": "Choose a share for the Restic backup data repository. This share is mapped to /data within the container.",
+                        "label": "Restic Data"                    
+                    }
+                },
+                "environment": {
+                    "PUID": {
+                        "description": "Enter a valid UID to run rest-server with. It must have full permissions to all Shares mapped in the previous step.",
+                        "label": "UID",
+                        "index": 1    
+                    },
+                    "PGID": {
+                        "description": "Enter a valid GID to use along with the same UID. It (or the above UID) must have full permissions to all Shares mapped in the previous step.",
+                        "label": "GID",
+                        "index": 2
+                    },
+                    "OPTIONS": {
+                        "description": "Enter rest-server specific options",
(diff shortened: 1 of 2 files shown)


Hooverdan96:
@ilium007, thanks for submitting this Rockon! We'll review it in the coming days.
One thing I've noticed at first glance though, if I understand this correctly, there is no UI involved in this setup, it's all via the container parameters or command line.
If that's correct, then you don't need to have the `ui` tag in the Rockon definition at all (since you also didn't tag any of the ports as a UI port).

ilium007:
Thanks, I'll remove that ui tag, it's all CLI.

Hooverdan96:
Hi @ilium007 since you indicated you were not continuing on Rockstor, wanted to ask whether you wanted to continue to pursue the addition of this Rockon? I've attempted some installs, and ran into a couple of hiccups there.

image

image

image

Summary page:
image

Installation proceeds and Rockon shows up as installed.

Checking `docker ps` doesn't show any undue restarts. `docker ps` shows these entries:

```
**WARNING** No user exists, please 'docker exec -it $CONTAINER_ID create_user'

Data directory: /data
Authentication disabled
Private repositories disabled
start server on [::]:8000
```

So, authentication has been disabled, but I still need to create a new user, which is easily done via:
```
docker exec -it rest-server create_user admin edgar
```
and it also shows the automatically created `.htpasswd` increases in size from 0 to 67 and 
`cat .htpasswd' yields the admin user name with an encrypted password.

Checking the `uid/gid` inside the container shows that it continues to run with as a root user. So, I think the UID/GID variables have no effect here, as they do not represent the `--user` docker option. In other cases (like the linuxserverio.io based images), these two variables will be parsed as part of the docker entrypoint script, and/or exposed variables, which for this docker image is not the case. So, those can be removed from the Rockon definition.

Are there other OPTIONS that can be maintained other than the `--no-auth` during the setup? I've tried various combos for defining a repository, or a password for the repository (with spaces, all of them together, or single ones), but in all cases I was getting the error messages back from the container logs that these options are unknown.

Do you have more insights into that? I do think that for this valuable Rockon, a write-up would probably really great for a use case to understand how this will work (since spinning up the Rockon is just the beginning. It seems that the official documentation does not provide great detail about the docker approach (vs. direct installation) and in the restic forum itself I only saw references to a possible "recipe" but that was not very helpful to get a better understanding of how to operated the docker version of the restic server ...

Just let me know, if you want to continue to collaborate on this or pause this if/when you might return to Rockstor?

I

Issue #447, opened 2025-07-18
Forgejo runner - more_info clarification


phillxnet:
A minor info/cosmetic follow-up re the more_info contents of the recently merged Forgejo runner rock-on in:

- Forgejo runner #444 #446

PR #446, opened 2025-07-15
Forgejo runner #444

phillxnet:
Fixes #444

### General information on project
This pull request proposes to add a new rock-on for the following project:
- name: Forgejo Actions runner
- website: https://docs.codeberg.org/ci/actions/
- description:  Self-hosted Forgejo actions runner.

### Information on docker image
- docker image: https://codeberg.org/phillxnet/forgejo-runner-wrapper
- is an official docker image available for this project?:  yes, but it is correctly minimalist; hence the layered helper image development.

Note: Adds a Forgejo runner Rock-on using a custom docker wrapper image `FROM:` the official code.forgejo.org/forgejo/runner upstream image.

### Checklist
- [x] Passes [JSONlint](https://jsonlint.com) validation
- [x] Entry added to `root.json` in _alphabetical_ order (for new rock-on only)
- [x] `"description"` object lists and links to the docker image used
- [ ] `"description"` object provides information on the image's particularities (advantage over another existing rock-on for the same project, for instance)
- [x] `"website"` object links to project's main website

Issue #444, opened 2025-07-02
Forgejo runner


phillxnet:
[Forgejo](https://forgejo.org/) runners, that run the acts in [Forgejo Actions](url), have an OCI image deployment method. It is proposed that we establish a Rock-on, with install config options sufficient to instantiate, for example, a 
[codeberg.org](https://codeberg.org/) remote runner.

- Official Container available from Forgejo's packages: https://code.forgejo.org/forgejo/-/packages
- Runner repo: https://code.forgejo.org/forgejo/runner
- Arch & OS: https://code.forgejo.org/forgejo/runner#architectures-os amd64 and arm64

See: **Forgejo Runner installation guide** https://forgejo.org/docs/latest/admin/runner-installation/

> The Forgejo Runner is a daemon that fetches workflows to run from a Forgejo instance, executes them, sends back with the logs and ultimately reports its success or failure.

- OCI image installation: https://forgejo.org/docs/latest/admin/runner-installation/#oci-image-installation

Auto config script:
```
forgejo-runner generate-config > config.yml
```
We could then edit this to container:

> container.docker_host = "automount"

# Pre-requisits

1. A new 'runner' must first be created in the Forgejo instance. See: [Standard registration](https://forgejo.org/docs/latest/admin/runner-installation/#standard-registration)
2. The remote runner must then execute: `forgejo-runner register` which must be supplied with:
  - the forgejo instance, i.e. "https://codeberg.org/"
  - a matching token from step 1.
  - a name for this runner.

PR #448, opened 2025-07-18
Forgejo runner - more_info clarification #447

phillxnet:
A minor info/cosmetic follow-up re the more_info contents of the recently merged Forgejo runner rock-on in:

- Forgejo runner #444 #446


phillxnet:
# Before

> # Additional information about Forgejo Runner Rock-on
> Intended as a Docker-in-Docker (dind) capable instance.
> Auto enables privileged mode, and automount re host docker service.
> 
> Auto configures: -v /var/run/docker.sock:/var/run/docker.sock

# After


> # Additional information about Forgejo Runner Rock-on
> Intended to shares the host docker instance.
> Auto enables privileged mode, and automount re host docker service.
> 
> Auto configures: -v /var/run/docker.sock:/var/run/docker.sock
> 
> Enabling sibling rather than child container creation for running workflows.

PR #450, opened 2025-08-02
Fixes #443

### LINKED ISSUE #443: MinIO admin port
From what I can tell this shouldn't actually be exposed:
https://github.com/rockstor/rockon-registry/blob/d7c3b0154bd71f4405a41540d8ef880d109a8d81/minio.json#L48-L51

Because changing it to match the actual external port changes what the server running within the container binds to, thus breaking the port forward. Instead this should always be `:9001`.

MikeMcPherson:
Fixes #443

Removes the option for the user to remap the MinIO API and console ports.  That functionality did not actually work.


phillxnet:
@MikeMcPherson Thanks for pressing on with this update/fix, as-always much appreciated.

But this PR looks to be adding a minio-dev.json, rather than modifying the existing [minio.json](https://github.com/rockstor/rockon-registry/blob/master/minio.json).

Have i miss-understood something here?

MikeMcPherson:
Should be OK now.

phillxnet:
@MikeMcPherson That looks more it :) . Thanks.

MikeMcPherson:
Cool!  Too many projects going at once.  Sorry for the oversight, and for the long delay in getting to this.

Hooverdan96:
I've tested the file in its current version:
image

image

image

After startup going to Admin UI using http://<ROCKSTOR_IP>:9001 which brings up the login:
image

image


So, that's all good.

@MikeMcPherson, since we're losing the UI button inside Rockstor (by moving the admin port to a fixed option the `ui` tag cannot be used), could you add something to the `description` (rather than the `more info` , so it's obvious), like this (or however you think it's best):

```
In order to access the built-in Admin Portal visit <code>http://ROCKSTOR_IP:9001</code>.
```
Once that's there, I think we can approve and merge.

Hooverdan96 (approved):
Looks good and test installation and startup is successful, with the ability to reach the Admin portal.

phillxnet:
PRODUCTION published

PR #451, opened 2025-08-09
Bareos backup server #94

### LINKED ISSUE #94: add a BareOS server rockon

phillxnet:
Fixes #94 

- name: Bareos
- website: https://www.bareos.com/
- description: see below header:

### Description
Network-based 'Backup and Recovery Open Source software' for all major operating systems. Director/Catalog/Storage and director-local (catalog/config backups) File services. Defaults to File based storage (a Share), and is intended as the Server-side to Client machines/DB's etc in need of backup services (Bareos Client software required).

#### Notes:
- Explicitly invoke `-u 105` during container run
- As a server-set, with ordered install, we can inherit (via the shared config arrangement) all intra Bareos service passwords. File and Storage containers are now able to self-config when uses in this intended AIO setup. This is akin to the rpms default behaviour of reading each-others config in order to auto-setup the initial server set.
-The SMTP_HOST environment variable are intended to enable access to Rockstor as an SMTP relay; enabling the default Bareos email notification with minimum setup.
- Docker network naming in Pascal Case. Exception is DB but this is a common Data Base abbreviation. We also drop File as "Fd" more closely matches the relevant container name.
- Contains link to a proposed Rock-on guide (https://github.com/rockstor/rockstor-doc/pull/546)

### Information on docker image/s
Multi-container Bareos server set Rock-on containing:
- Director Service (overall controller): https://codeberg.org/phillxnet/-/packages/container/bareos-director/latest
- Catalog Service (Postgres DB for Director): Official Postgres 16 image.
- Storage Service (Back-end Storage for Director and File (client) services): https://codeberg.org/phillxnet/-/packages/container/bareos-storage/latest
- Director-local File (client) Service (for Catalog backup capability): https://codeberg.org/phillxnet/-/packages/container/bareos-file/latest
- Web-UI Service (Director Web-UI interface): https://codeberg.org/phillxnet/-/packages/container/bareos-webui/latest

- is an official docker image available for this project?: No.

### Checklist

### DIFF
--- a/bareos-backup-server.json
+++ b/bareos-backup-server.json
@@ -1,0 +1,208 @@
+{
+  "Bareos Backup Server": {
+    "description": "Network-based '<strong>Ba</strong>ckup and <strong>Re</strong>covery <strong>O</strong>pen <strong>S</strong>ource software' for all major operating systems. Director/Catalog/Storage and director-local (catalog/config backups) File services. <a href='https://rockstor.com/docs/interface/docker-based-rock-ons/bareos-backup-server.html' target='_blank'>Rock-on guide</a>.<p><strong>Before install run: </strong><code>groupadd --system --gid 105 bareos</code> && <code>useradd --system --uid 105 --comment 'bareos' --home-dir /var/lib/bareos -g bareos -G disk,tape --shell /bin/false bareos</code><p><strong>After install requirement:</strong> click 'i' icon for instructions.<p>Based on <a href='https://codeberg.org/phillxnet/-/packages' target='_blank'>custom docker images</a> from the Rockstor maintainers, available for amd64 architecture only.</p>",
+    "more_info": "<h4>Required post-install configuration</h4><ol><li>System - Network - Add Connection (Name=BareosFdToStorage Type=docker).</li><li>This Rock-on's spanner icon - Networking - Rocknets - select BareosFdToStorage for bareos-fd & bareos-storage.</li><ol>",
+    "ui": {
+      "slug": "bareos-webui"
+    },
+    "version": "0.0.1",
+    "website": "https://www.bareos.org/",
+    "container_links": {
+      "bareos-dir": [
+        {
+          "name": "BareosDirToDB",
+          "source_container": "bareos-db"
+        },
+        {
+          "name": "BareosDirToStorage",
+          "source_container": "bareos-storage"
+        },
+        {
+          "name": "BareosDirToFd",
+          "source_container": "bareos-fd"
+        },
+        {
+          "name": "BareosDirToWebui",
+          "source_container": "bareos-webui"
+        }
+      ]
+    },
+    "containers": {
+      "bareos-db": {
+        "image": "postgres",
+        "tag": "16",
+        "launch_order": 1,
+        "volumes": {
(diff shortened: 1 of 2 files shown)


FroggyFlox (approved):
What a great Rock-On, @phillxnet!
That is definitely an advanced use one but I could easily follow the associated write-up you prepared and get a working backup of my laptop with it.

Hooverdan96:
One quick question, on the environment variables, since you're not using the `<index>` tag to sort/group them in a particular order, you think it matters? Or the sequence they show up makes the most sense already? Other than that, this is a great addition, and a good example of how a complex Rockon can be implemented (even though it required you to roll your own containers 😄 )

phillxnet:
@FroggyFlox @Hooverdan96 Cheers, I'll get to the merging & publishing of this shortly then.

I was hoping for it to be a little less complex over-all. But we do have some rockstor-core improvements that would help on that front also. So all in good time hopefully. Incidentally, on that front, I think for example we could have auto-generated (random) passwords. But that can be a rockstor-core issue discussion.

@Hooverdan96 

> since you're not using the `<index>` tag to sort/group them in a particular order, you think it matters? Or the sequence they show up makes the most sense already?

I'd forgot about that actually; thanks. But as-is they appear to be fairly workable so I'll go with the merge and we can re-visit as needed in some follow-up PR as and when I think.

Re:

> (even though it required you to roll your own containers 😄 )

Yes indeed :) , originally I did try to use the only example of such containers, as indicated in the container readmes, and earlier drafts of this PR, but they were not suited to our purpose and after much persevering, and bug finding/fixing (some already reported and not applied) and some major doc lag issues I felt the best course of action was to roll-our-own. At least then we are using official packages directly from upstream (on a supported OS), which the only other container image examples did not use; on both accounts. So over-all I'm happy with this path, but yes, I/we do now have a new set of puppies!!! However we have already had community input/feedback on at least the bareos-webui so that's good (https://codeberg.org/phillxnet/bareos-webui/issues/5). Plus @FroggyFlox has found a '#' password corner case that we can tend to so all good I think. Plus I'm feeling a little more up on container technologies and best practice now, something I also feel happier about. So all good all around hopefully.

PR #461, opened 2025-09-04
Update Jetbrains youtrack image version (#455)

### LINKED ISSUE #455: Update Jetbrains You Track to a more recent version
The current You track docker image us using the underlying version: jetbrains/youtrack:2019.2.58831 which is over 5 years old. 

The product is still actively maintained, current version is 2025.2.94372:

https://hub.docker.com/r/jetbrains/youtrack

The recommendation is to move to the most recent version, unless there is a compelling reason to stay on this one, or pick a different one.

Hooverdan96:
Fixes #455  .

Update jetbrains youtrack version and clean up some texts

### Checklist


phillxnet (dismissed):
@Hooverdan96 Thanks for the update here, the associated image was some-what aged. Much appreciated.

Further to @FroggyFlox approval also here, this looks to be working as intended.

After having expanded the test VM to 3 GB RAM and doing a fresh install again to avoid the already commented OOM killer issues I was able to Web-UI login (using the admin user created on initial YouTrack Web-UI setup and receive the Demo into.

A note however on another normalisation that I've just added to our ongoing list of standards for Rock-ons. We already have examples of Rock-ons that require specific user:group, so we should re-visit this one which shares that requriement. I.e. we can specify " (13001:13001)." at the end of all Shares here. Not a show-stopper thought as we can revisit as we get time.

youtrack-token-retrieval

WebUI-first-visit-token-entry

Setup-or-Upgrade

Initial-demo_project-login

My test install used the provided free up-to 10 user license presented during the install.

phillxnet:
@FroggyFlox as per @Hooverdan96 https://github.com/rockstor/rockon-registry/pull/461#issuecomment-3255801344 I would also rather you second these changes as I know you were dog-fooding this Rock-on and I am essentially unfamiliar with this product. But I did follow you doc to accomplish a relatively smooth install; post moving to a 3 GB RAM VM instance.

As such I'll hold off on merging until we have your input. But it does look to be installing and starting as indicated by my above review.

FroggyFlox:
Thank you to both of you for all that work! 
I unfortunately have not been using that product either so my knowledge of it is only remote memories at this point, I'm afraid. 

Sorry I can't help further on this. It does look really nice and ready for merging, though, especially with all your testing.

Hooverdan96:
I've updated the share ownership piece - I felt it needed to call out what the UID:GID requirement refers to. If that's too verbose I can scale it back to the UID:GID.

One other thing to point out (and it has been discussed before) the reference to the installation instruction has been (and still is) under the more_info tag that is only available up after the installation is complete. This is not really ideal. Maybe moving the link to the documentation into the main description instead? I understand the desire to not overwhelm the "front" of the Rockon with too much information, but would feel that for all Rockons that have documentation on the website the reference should show up in the Rockon list, and not after the installation.

Here is the related issue from the core repo [#2231](https://github.com/rockstor/rockstor-core/issues/2231)

phillxnet (approved):
@Hooverdan96 Thanks for the follow-up commit re 6ad676700ac2db3e2e4916c17c8468fe87b1fe63 "clarify share ownership requirements", that's a definite improvement.  Not too verbose either. We can do a follow-up PR on this and most other Rock-ons as the details I proposed for indicating this were purposefully detailed as such to aid our future machine reading of this type of info. I'm hoping to prepare all Rock-ons, over time, to be ameanable to parsing this info from what will be a legacy Rock-on format. Hence trying to be strict with formatting for these new additions.

Post having eased our front end development, we can transition to abstracting this info from legacy Rock-ons (our current format) so that such info is independently displayed. And can be enacted on by an auto-create of all Shares for example. So having a strict format now will help our transition to later versions of the Rock-on format.

But whatever - this is definitely a nice move in the right direction; thanks for squeezing this one in. Later we can express all such definitions into the last element in a Description within brackets to ease syntactic extraction from our current (and future legacy) Rock-on definition format.

Approving and moving to merge so we can get this key update out. Thanks for your engagement on all this maintenance and for suffering all these cosmetic tweaks, I'll try to get some similar work in myself soon in this repo as we do now have quite a number of Rock-ons to herd.

phillxnet:
PRODUCTION published

PR #482, opened 2025-11-13
added n8n rock-on config

BrokenOnedroid:
# Adds n8n

- name: n8n
- website: https://n8n.io
- description: n8n lets you visually automate any task or service with drag‑and‑drop workflows, no code required.

- docker image: https://hub.docker.com/r/n8nio/n8n
- is an official docker image available for this project?: This uss the official image

### Checklist

### DIFF
--- a/n8n.json
+++ b/n8n.json
@@ -16,7 +16,7 @@
         "volumes": {
           "/home/node/.n8n": {
             "description": "Persistent n8n data (workflows, credentials, settings)",
-            "label": "n8n Data"
+            "label": "n8n Data [e.g. n8n-data]"
           }
         },
         "opts": [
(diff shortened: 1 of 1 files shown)


Hooverdan96:
@BrokenOnedroid thanks for bringing this Rockon. I've tried a quick installation to date, but I seem to not be able to keep running (and so of course, the container logs are not accessible while it keeps restarting. So, I have to further look into it.

The options you've outlined (e.g. the webhook), are they required for the standalone version or only if it's run in conjunction with Traefik (like you've mentioned in the Rockstor [forum](https://forum.rockstor.com/t/n8n-rock-on-config/10786))?


As a side note, and @phillxnet brought this up in a more recent PR that was submitted:

We are trying to make some more options more explicit and move them from the descriptions (which are hidden within a tooltip) directly to the label. One example is the volumes object:

https://github.com/rockstor/rockon-registry?tab=readme-ov-file#volumes-object

@phillxnet's Bareos Backup Server Rockon is a good example:
https://github.com/rockstor/rockon-registry/blob/master/bareos-backup-server.json

BrokenOnedroid:
Okay, I'm going to take a look at the mentions. And try to replicate the error.

BrokenOnedroid:
If i take the changed n8n and change the owner of the share to one of the user. 
The container starts. 
I can access the webui.

image

image

n8n.json
```diff
@@ -0,0 +1,44 @@
...
+          ["-e", "DB_SQLITE_POOL_SIZE=4"],
+          ["-e", "N8N_RUNNERS_ENABLED=true"],
+          ["-e", "N8N_DIAGNOSTICS_ENABLED=false"],
+          ["-e", "N8N_VERSION_NOTIFICATIONS_ENABLED=false"],
+          ["-e", "N8N_GIT_NODE_DISABLE_BARE_REPOS=true"],
+          ["-e", "N8N_SECURE_COOKIE=false"]
+         ]
+      }
+    },
+    "description": "n8n is a workflow automation tool. It allows you to connect apps and services, automate tasks, and create complex workflows with a visual editor. Available for amd64 amd arm architecture",
```
Hooverdan96:
```suggestion
    "description": "n8n is a workflow automation tool. It allows you to connect apps and services, automate tasks, and create complex workflows with a visual editor.<p>Based on the <a href='https://hub.docker.com/r/n8nio/n8n' target='_blank'>official docker image</a>, available for amd64 amd arm architecture",
```

Adding the docker image link in here.

Hooverdan96 (changes requested):
See two suggestions I've made, one to clarify the share permissions required, one to add the used docker image link to the description.

phillxnet:
> I am suggesting for this case (at least until the introducion of the --uid with both user and group in the parameters) to put an explicit share requirement into the volume description. Let us know whether you disagree with that approach.

Agreed. This does seem like the way to go. Linking below to what will likely be our first extension to our existing --user interface:

- https://github.com/rockstor/rockstor-core/issues/3029

phillxnet:
@BrokenOnedroid Apologies for the now required rebase and ideally a squash of all your commits on this pull request. We are, bit by bit, normalising our Rock-on format and this has now been done on the root.json. Plus, ideally, the initial presentation pre-review, should have only a single commit. Note also the suggestions proposed by @Hooverdan96.

PR #493, opened 2025-12-19
Replace minio docker image (#481)

### LINKED ISSUE #481: MINIO community edition changing to build-from-source
Rockstor forum user @emptyNAS highlighted a change that has apparently occurred with the Minio project:

https://forum.rockstor.com/t/minio-no-longer-updates-the-docker-image/10785

With the security release on 10-15-2025 the decision to become a source only distribution for the community edition:

https://github.com/minio/minio/commit/9e49d5e7a648f00e26f2246f4dc28e6b07f8c84a

which in turn also seems to affect the docker container generation that the project used to provide. Very unfortunate.

The currently used image mimio/mimio could possibly be replaced by a regularly built alternative, e.g.

https://hub.docker.com/r/cleanstart/minio

which seems to have the latest release incorporated in its image. Have to evaluate whether the same parameters are exposed as in the native image.

Hooverdan96:
Fixes #481 .

Unfortunately, the MinIO project is not providing an updated official container anymore. This PR proposes the use of a custom docker image that is regularly built from source on top of a security hardened base image. This image change is a drop-in, as the same parameters are used for the custom docker image.

- docker image: https://hub.docker.com/r/cleanstart/minio/
- is an official docker image available for this project?: yes, but not updated anymore (see above)

### Checklist

### DIFF
--- a/minio.json
+++ b/minio.json
@@ -17,7 +17,7 @@
         },
         "volumes": {
           "/data": {
-            "description": "Choose a share for MinIO object storage",
+            "description": "Choose a share for MinIO object storage that can be accessed by the Rock-on's user [e.g. create a group clngroup with GID=102 and create a user clnstrt with UID=180802 assigned to the previously created group].",
             "label": "Object Storage [e.g. minio-obj-store]"
           }
         },


MikeMcPherson:
I don't know anything about CleanStart, but if you say it is good them I'm fine with it.

Hooverdan96:
I read a bit about the company offering it ... but would not be qualified to say whether they are terrible or outstanding. Their image size appears fairly small and at least they claim they're very diligent on the security hardening side. And, finally it would be just a drop-in replacement (including the `--console-address` parameter)

Of course the Bitnami one seems to be the second most popular one after the minio image, though it also has not been updated in a few months so not sure it's a better choice?
This would likely require a change away from the console parameters (I assume one would use the environment parameter `MINIO_CONSOLE_PORT_NUMBER` instead?

MikeMcPherson:
Another possibility would be to drop MinIO and move to RustFS.  It is S3-compatible, apparently runs circles around MinIO performance-wise, and is distributed under the simpler Apache License.

Hooverdan96:
That is a good point on the open source version development continuation. I didn't see that part.
So, that might be the better alternative, especially since it's under a lot of development currently.

@MikeMcPherson, since you were the original proponent of minio, would you favor that approach?
Or continue with the drop- in (I just want to make sure we don't keep an "aged" version) and introduce the RustFS as a new one/alternative? This could give other minio users both opportunity as well as a time window to transition to RustFS if they so desire.

PR #532, opened 2026-05-05
Sunset ecodms (#457)

### LINKED ISSUE #457: ecoDMS archival software outdated
It looks like the AllInOne product from ecoDMS has not been updated in  7 years:

https://hub.docker.com/u/ecodms

In the meantime new versions have come and gone (last one around 2 years ago). But all of those seem to have been superseded by https://hub.docker.com/r/ecodms/ecodms

this should probably be updated to the latest version tree.

Hooverdan96:
Fixes #457  .

As suggested, remove ecoDMS from the Rockon registry due to a very outdated container image version, indicating that it might not be currently used by any of the Rockstor user base (since, by design, the actual download/installation numbers cannot be tracked).


phillxnet (approved):
@Hooverdan96 Thanks for seeing to this sunsetting. Much appreciated.

With a local production like setup Rock-on host this looks to behave as expected. I.e. no ecoDMS. And not log issued indicated.

Moving to merge and publish.

phillxnet:
PRODUCTION published.

PR #534, opened 2026-09-11
Seerr

sschwetz:
> You can find guidelines on which OCI image to use in the [Rockstor's documentation](http://rockstor.com/docs/contribute_rockons.html#which-docker-image-should-i-use), 

Fixes # N/AS

- name: Seerr
- website: https://seerr.dev
- description: Seerr is a media management framework that integrates with Plex/Jellyfin and integrates with sonarr and radarr to allow end users of these platforms to perform media requests.

### Information on OCI image
- OCI image: ghcr.io/seerr-team/seerr
- is an official OCI image available for this project?: yes

### Checklist
- [X] `"description"` object lists and links to the OCI image used


phillxnet:
@sschwetz This looks like a nice project.

I've just had a quick look at the following https://docs.seerr.dev/getting-started/docker/ to check mappings etc and it looks like your  `/config` volume mapping may be in error. The doc link looks to be indicating this should be `/app/config `. There is also no mention of PUID, or PGID as available variables. We do now have options to set these within our testing variant which is soon to be promoted to stable. Check out the README.md here in this repo for available options on that front. Your suggested user:group (in brackets) for any mapped shares would have to indicate a must be 1000:1000 type arrangement then though of course. Given the Rock-on would then be run as a hard-wired user:group in that case. Their above doc indicates the docker image is able to entertain that so that's good.

The PUID and PGID nomenclature originates, or was popularised, from/by linuxserver docker image definitions. But is not shown as an available environmental variable for this docker image.

You may want to mark this proposed pull request as a draft here on GitHub if it is still, as it seems, in development. That way it can server for folks to chip-in with advice and such.

Incidentally, we already set every Rock-on's timezone to that defined within the Rockstor install itself. Presumably their docker image will default to that if the TZ environmental variable is not set.

sschwetz:
@phillxnet I have updated the environment setup and can confirm that seer does not support setting the UID:GID. This file was originally using the overseer project, and I missed that the support to set the UID and GUID had changed. I can also confirm that this new image used /app/config instead of /config.

As you are setting the timezone, I have removed the explicit setting.

I have destroyed and redeployed the container a couple of times, and the config is successfully being reloaded in the new container.

Thanks for checking this out

phillxnet:
@sschwetz Thanks for looking into these elements. Always tricky to track project changes.

I see you have also elected to go for the hard-wired user:group approach in your last change re:

> Must be owned by the UID:GID as (1000:1000)

Can you also confirm that this is the default user that is adopted by this docker image. I ask as we run, as docker does, as root:root all containers, unless the Rock-on uses our uid & gid definitions.

When I looked at: https://docs.seerr.dev/getting-started/docker/

The **Docker CLI** tab has: 

> To run the container as a specific user/group, you may optionally add --user=[ user | user:group | uid | uid:gid | user:gid | uid:group ] to the above command.

suggesting they are able to run as non root user:group, which many docker images cannot handle successfully as they depend on initial root privaledges in order to change the user they run as later on, i.e. via environmental variables for example. But of course it is always better to have never been granted root in the first place.

- our uid & gid definitions essentially wrap the docker "--user" option referened there. See our following README.md entry in this repo: https://github.com/rockstor/rockon-registry#uid--gid-elements for the options available. We are also able to key from a Share's owner:group - but given they are, by default, root:root, this can end up exposing our less technical users to inadvertently running containers as the root user. Something partially approached by containers that drop initial root privaledges after using them to change to a set, or defined user.

Curiously thought, under their **Docker Compose** section, same doc, I see:

```
# Chown the folder as the container runs as the `node` user (UID 1000).
chown -R 1000:1000 /path/to/appdata/config
```

So it is unclear if this image initially requires root to then change to 1000:1000, in which case we now have that covered as that is dockers, and our default (without our uid & gid elements), and you have specified a hard requirement to set the Share owner:group appropriately. Though note that there is, by default, no (named) group 1000 in opensuse. But the ownership alone may server here.

I.e. take a look at the owner:group of the config files written by this latest version of your Rock-on to see what the Rock-on is actually 'writing' as. I.e. does it in fact change it's user:group down from root to 1000:1000. This is all good to know before we entertain a final reivew. Also, have you now used this latest version successfully? I.e. does it install, uninstall, autostart when enabled on boot. Pick-up where it was left when turned off and on again. That type of think.

Thanks for reporting on the context of this Rock-ons development. Also good to know. All these moving goal posts makes this kind of development tricky.

And thanks for taking all this effort to submit what looks like a popular project to make it accessible to our own project's community.

A more recent effort on the Rock-ons has tried to normalise their definitions. So a look over other examples of Rockon in the latest version of our repo can help with conforming submission to what we hopefully now have established as a norm. We also now have the following in-development tool that can help with this conformance:

- Rockon-validator: https://github.com/rockstor/rockon-validator

Hope this helps. And thanks again for your patience & persistence in getting this Rock-on up the hill of initial submission.

---

I'm assuming also that you are aware of the following test mechanism to prove function within Rockstor:

- https://rockstor.com/docs/contribute/contribute_rockons.html

I.e. establishing a local directory on your Rockstor instance  so that the proposed Rock-on json definition can be proven in-place prior to submission on the public Rock-on repo.

PR #537, opened 2026-09-18
add copyparty

### LINKED ISSUE #536: add copyparty
will be adding a copyparty rockon.

kanecko:
> You can find guidelines on which OCI image to use in the [Rockstor's documentation](http://rockstor.com/docs/contribute_rockons.html#which-docker-image-should-i-use), 

Fixes #536 .

- name: Copyparty
- website: https://github.com/9001/copyparty
- description: It's a lightweight file sharing website, akin to filebrowser. It is simple to use until you need to click on a button or find a configuration setting.

### Information on OCI image
- OCI image: https://hub.docker.com/r/copyparty/ac
- is an official OCI image available for this project?:  yes

### Checklist
- [x] `"description"` object lists and links to the OCI image used


copyparty.json
```diff
@@ -0,0 +1,45 @@
+{
+  "Copyparty": {
```
kanecko, line 2: according to the owner of the project it should be called either "copyparty" or "Copyparty". As one word.

copyparty.json
```diff
@@ -0,0 +1,45 @@
+{
+  "Copyparty": {
+    "description": "Copyparty is a lightweight file sharing hub with accelerated resumable uploads, deduplication, WebDAV, zeroconf, a media indexer, image/audio/video thumbnails and audio transcoding. It is served over a single HTTP port and needs no client software beyond a web browser.<p>The container runs as the owner and group of the config Share, so that account must have read and write access to every Share mapped into the container.</p><p>Note: You must add a config file (copyparty.conf) to your copyparty Config [e.g. copyparty-config] Share. <a href='https://github.com/9001/copyparty/blob/hovudstraum/docs/examples/docker/basic-docker-compose/copyparty.conf' target='_blank'>Example</a>.</p><p>Based on the <a href='https://hub.docker.com/r/copyparty/ac' target='_blank'>official OCI image</a>, available for amd64 and arm64 architecture.</p>",
+    "version": "1.0",
+    "website": "https://github.com/9001/copyparty",
+    "icon": "https://raw.githubusercontent.com/9001/copyparty/hovudstraum/docs/logo.svg",
+    "opts": [
```
kanecko, line 7: rockon validator complained here:
```
-    "opts": [
-        ["--stop-timeout", "15"]
-    ],

```

but i left it in because the owner of the project said the thumbnail generator should have a grace period, IIRC.

copyparty.json
```diff
@@ -0,0 +1,45 @@
...
+        ["--stop-timeout", "15"]
+    ],
+    "containers": {
+      "copyparty": {
+        "image": "copyparty/ac",
+        "tag": "latest",
+        "launch_order": 1,
+        "uid": -1,
+        "gid": -1,
```
kanecko, line 16: I have tested it on Rockstor V5.1, by first creating the shares, then the "copyparty" user and group, then i assigned uid/gid to the share.
I noticed that when I upload a file, it will be given the "root" group instead of the expected "copyparty".
Please test this on V5.5+. I'm not sure if it's a V5.5 feature or I misconfigured it.

kanecko:
to test it, you also need to add a some.conf file into the config share.
this is the config i used:
```
# not actually YAML but lets pretend:
# -*- mode: yaml -*-
# vim: ft=yaml:


[global]
  hist: /cfg/hists/  # store internal .hist folders in the cfg volume
  e2dsa  # enable file indexing and filesystem scanning
  e2ts   # enable multimedia indexing
#  ansi   # enable colors in log messages (both in logfiles and stdout)

  # q, lo: /cfg/log/%Y-%m%d.log   # log to file instead of docker

  # p: 3939          # listen on another port
  # ipa: 10.89.      # only allow connections from 10.89.*
  # ipa: 172.16.4.0/23  # ...or only 172.16.4.* and 172.16.5.*
  ipa: lan         # ...or allow LAN only; reject internet IPs
  df: 1           # stop accepting uploads if less than 1 GB free disk space
  #ver              # show copyparty version in the controlpanel
  # grid             # show thumbnails/grid-view by default
  #theme: 2         # monokai
  # name: datasaver  # change the server-name that's displayed in the browser
  # stats, nos-dup   # enable the prometheus endpoint, but disable the dupes counter (too slow)
  # no-robots, force-js  # make it harder for search engines to read your server

  # enable version-checking by uncommenting one of the vc-url lines below;
  # shows a warning-banner in the controlpanel if your version has a known vulnerability
  #vc-url: https://api.github.com/repos/9001/copyparty/security-advisories?per_page=9
  vc-url: https://api.copyparty.eu/advisories
  #vc-exit  # panic and shutdown instead of just showing the warning


[accounts]
  hello: operator  # username: password


[/]            # create a volume at "/" (the webroot), which will
  /w           # share /w (the docker data volume)
  accs:
    r: *      # everyone gets read-write access, but
#    rwmda: ed  # the user "ed" gets read-write-move-delete-admin
#  flags:
#    e2ds       # enable filesystem-scanning for this volume only
```

copyparty.json
phillxnet: @kanecko You are correct in your guess that the gid element is not yet in our last (V5.1) stable. The following README.md section in this repo has the version details we need here:
- https://github.com/rockstor/rockon-registry#uid--gid-elements

> The gid element depends on the existence of a uid element and requires Rockstor 5.5.0-0 onwards.

phillxnet:
@kanecko This is a very thorough presentation, thanks. For Rock-ons that must have set share owner:group you could look to our "Bareos Backup Server" for how we word the description re owner:group i.e. (bareos:bareos) at the end of the volume description where it applies. 

In time we hope to have a more automated installer that can key off such info, hence the formatting normalization we did recently in this repo. And can also, in time, extract such info in later UI developments to make things clearer as we go.

kanecko:
added the  (copyparty:copyparty) wording.
added "Rockstor V5.5.0 or later." since it is dependent upon the "gid" feature.

